PR #4, opened 2024-01-31
[ Álvaro Zamanillo Sáez ] iP

azamanis:


src/main/java/Deadline.java
```diff
@@ -0,0 +1,18 @@
+public class Deadline extends Task{
+    protected String by;
+    public static final String[] args = {"/by"};
```
Hws2209: This is a constant? Should be capitalised?

src/main/java/Kvothe.java
```diff
@@ -0,0 +1,204 @@
+import java.util.Scanner;
+
+public class Kvothe {
+
+    private static final int MAXTASKS = 100;
```
Hws2209: I like that this is set as a constant

src/main/java/Event.java
```diff
@@ -0,0 +1,24 @@
...
+    protected String to;
+
+    public static final String[] args = {"/from", "/to"};
+    public Event(String description, String from, String to){
+        super(description);
+        this.from = from;
+        this.to = to;
+    }
+
+    public String getFrom(){
```
Hws2209: Need a space in front of the curly braces?

src/main/java/Kvothe.java
```diff
@@ -0,0 +1,204 @@
...
+    //TODO: move to Task class?
+    //TODO: Check that the arguments are the correct ones
+    private static String[] parseLine(String line, String[] args) {
+        String[] lineWords = line.split(" ");
+        String[] newArgs = new String[args.length + 1]; // description + args
+
+        int lastWordIndex = 1;
+        int argsIndex = 0;
+
+        // Get the description
```
Hws2209: Obvious comment not needed?

src/main/java/Kvothe.java
```diff
@@ -0,0 +1,204 @@
...
+
+    private static void add(String line) {
+
+        String lineWords[] = line.split(" ");
+        String command = lineWords[0];
+        Task newTask = null;
+        String args[];
+
+        switch (command) {
+            case "todo":
```
Hws2209: Should have no indentation for case clauses

Hws2209 (commented):
Some small changes for coding standards

src/main/java/Event.java
```diff
@@ -0,0 +1,24 @@
...
+        return this.from;
+    }
+
+    public String getTo(){
+        return this.to;
+    }
+
+    @Override
+    public String toString(){
```
claribelho: Method name could be clearer

src/main/java/Kvothe.java
claribelho: Underscore should be used to separate words for this constant

src/main/java/Kvothe.java
```diff
@@ -0,0 +1,204 @@
...
+        String line = in.nextLine();
+
+        while (!line.equals(stopWord)) {
+            String lineWords[] = line.split(" ");
+            String command = lineWords[0];
+
+            int taskNumber;
+
+            switch (command) {
+                case "list":
```
claribelho: There shouldn't be indentation for the case clauses in switch

src/main/java/Todo.java
```diff
@@ -0,0 +1,13 @@
...
+        public static final String[] args = {};
+
+        public Todo (String description){
+            super(description);
+        }
+
+        @Override
+        public String toString(){
+            return "[T]" + super.toString();
+        }
```
claribelho: Basic indentation should be 4 spaces :)

src/main/java/Kvothe.java
```diff
@@ -0,0 +1,204 @@
...
+    private static void echo(String line) {
+        System.out.println("\t\t____________________________________________________________");
+        System.out.println("\t\t" + line);
+        System.out.println("\t\t____________________________________________________________");
+
+    }
+
+    //TODO: move to Task class?
+    //TODO: Check that the arguments are the correct ones
+    private static String[] parseLine(String line, String[] args) {
```
fungg0: This method has about 40 lines of code (>30 LOC). I think you can shorten this method.

src/main/java/Kvothe.java
```diff
@@ -0,0 +1,204 @@
...
+        String greeting = "Hello! I'm Kvothe.\n"
+                + "\t\tWhat can I do for you?\n";
+
+        String bye = "Bye. Hope to see you again soon!";
+
+        echo(greeting);
+
+        interact();
+
+        echo(bye);
```
fungg0: There are 2 different levels of abstraction here. You can consider making a new method for the string declaration.

src/main/java/Kvothe.java
```diff
@@ -0,0 +1,204 @@
...
+        tasks[index - 1].setIsDone(false);
+        echo("OK, I've marked this task as not done yet:\n\t\t" + tasks[index - 1]);
+    }
+
+    private static int getTaskNumber(String[] args){
+
+        int index = 0;
+        if (args.length < 2 || args.length > 2) {
+           echo("Sorry. Please specify one task number.");
+           return -1;
```
fungg0: This seems like a magic number. Perhaps you could represent the -1 as a constant named INVALID_INDEX.

src/main/java/Kvothe.java
```diff
@@ -0,0 +1,204 @@
...
+    private static String[] parseLine(String line, String[] args) {
+        String[] lineWords = line.split(" ");
+        String[] newArgs = new String[args.length + 1]; // description + args
+
+        int lastWordIndex = 1;
+        int argsIndex = 0;
+
+        // Get the description
+        String description = "";
+        while (lastWordIndex < lineWords.length && !lineWords[lastWordIndex].startsWith("/")) {
```
fungg0: I think this expression is a little complicated. I think it will be more intuitive if the code is broken down into more steps.

fungg0 (commented):
Your code could use some fixes, but otherwise it is clean!

PR #8, opened 2024-02-01
[Wong Quan Wei Alvin] iP

alvinnnnnnnnnn:


src/main/java/Wongster.java
```diff
@@ -0,0 +1,129 @@
...
+        this.isDone = false;
+    }
+
+    public String getDescription() {
+        return description;
+    }
+}
+
+
+class ToDo extends Task {
```
claribelho: new classes should be created in a new file! same for the other classes :)

src/main/java/Wongster.java
```diff
@@ -0,0 +1,129 @@
+import java.util.Scanner;
+import java.util.Arrays;
+class Task {
+    protected String description;
+    protected boolean isDone;
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+    public String getStatusIcon() {
```
claribelho: I like how your method names are easy to read and understand!

src/main/java/Wongster.java
```diff
@@ -0,0 +1,129 @@
...
+
+public class Wongster {
+    public static void main(String[] args) {
+        String name = "Wongster";
+        System.out.println("Hello! I'm " + name + "\nWhat can I do for you?\n");
+        echo();
+    }
+
+    public static void echo() {
+        Task[] userList = new Task[100];
```
claribelho: magic number! should be declared as a static final variable

src/main/java/Wongster.java
```diff
@@ -0,0 +1,129 @@
...
+                System.out.println(userList[index].getStatusIcon() + " " + userList[index].getDescription());
+            } else if (userInput.equalsIgnoreCase("list")) {
+                if (userListItems == 0) {
+                    System.out.println("There are no tasks in your list. Please input at least one!");
+                } else {
+                    System.out.println("Here are tasks in your list: \n");
+                    for (int i = 0; i < userListItems; i++) {
+                        System.out.println((i + 1) + "." + userList[i]);
+                    }
+                }
```
claribelho: Perhaps a little too nested?

src/main/java/Wongster.java
```diff
@@ -0,0 +1,129 @@
...
+        String name = "Wongster";
+        System.out.println("Hello! I'm " + name + "\nWhat can I do for you?\n");
+        echo();
+    }
+
+    public static void echo() {
+        Task[] userList = new Task[100];
+        int userListItems = 0;
+        Scanner scanner = new Scanner(System.in);
+        while (true) {
```
claribelho: while loop is very long... perhaps could create methods that can be called?

claribelho (commented):
Only has one file, making it a little hard to read :o but did use inheritance of Task class!

src/main/java/Deadline.java
```diff
@@ -0,0 +1,10 @@
+public class Deadline extends Task {
+    protected String by;
+    public Deadline(String description, String by) {
+        super(description);
+        this.by = by;
+    }
+    public String toString() {
```
okkhoy: need some vertical spacing

src/main/java/Wongster.java
```diff
@@ -0,0 +1,101 @@
...
+                System.out.println("Noted. I've removed this task:\n" + " " + currentTask +
+                        "\nNow you have " + Integer.toString((userListItems)) +" tasks " + "in the list.\n");
+            } else {
+                System.out.println("Please input a a proper task!");
+                System.out.println("Help: Tasks start with todo, deadline or event.");
+            }
+        }
+        scanner.close();
+    }
+}
```
okkhoy: main method seems to do a lot of work

okkhoy (commented):
Not much code to review.

PR #17, opened 2024-02-02
[Damien Wee] iP

damiwee:


src/main/java/BossMan.java
```diff
@@ -0,0 +1,147 @@
...
+        default:
+            System.out.println("Unknown command\n" + SEP);
+        }
+    }
+
+    private String[] parseUserInput(String userInput) {
+        return userInput.trim().split("\\s+", 2);
+    }
+
+    private void handleMarkCommand(String commandArgs) {
```
sweijie24: The handleMarkCommand and handleUnmarkCommand seems to be handled similarly. Consider refactoring it into a separate method to avoid duplication

src/main/java/TaskList.java
```diff
@@ -0,0 +1,62 @@
+import java.util.ArrayList;
+import java.util.List;
+
+public class TaskList {
+    private final String SEP = "____________________________________________________________";
```
sweijie24: Consider making the constant here "static" if it is not being modified across all instances

src/main/java/TaskList.java
```diff
@@ -0,0 +1,62 @@
...
+        System.out.println(SEP);
+    }
+
+    public void markTask(int num)  {
+        if (isValidTask(num)) {
+            this.TASKS.get(num - 1).setMark();
+            System.out.println("Nice! I've marked this task as done:");
+            echo(num);
+        } else {
+            System.out.println("Selected index out of range" + "\n" + SEP);
```
sweijie24: Consider also throwing an exception instead of just printing a message

src/main/java/TaskList.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    public void addTask(Task task) {
+        TASKS.add(task);
+    }
+
+    public void printTasks() {
+        System.out.println("Todo List:");
+        int counter = 1;
+        for (Task t: TASKS) {
```
sweijie24: Consider using a traditional for-loop standard since you are using a counter anyways. This can help the code to be more readable and perhaps avoid further issues in the event of any future modifications

src/main/java/TaskList.java
```diff
@@ -0,0 +1,62 @@
...
+            System.out.print(counter + ". ");
+            t.printTask();
+            counter += 1;
+            System.out.println();
+        }
+        System.out.println(SEP);
+    }
+
+    public void markTask(int num)  {
```
sweijie24: Similarly, the markTask and unmarkTask can perhaps be refactored into 1 whole method

src/main/java/BossMan.java
```diff
@@ -0,0 +1,147 @@
...
+    private final TaskList TASK_LIST;
+
+    public BossMan() {
+        this.SCANNER = new Scanner(System.in);
+        this.TASK_LIST = new TaskList();
+    }
+
+    public void greetUser() {
+        System.out.println(SEP + "\nHello! I'm BossMan");
+        System.out.println("What can I do for you?\n" + SEP);
```
royaliceblue: Would it be better to print a line break using the System.lineSeparator() as the behavior of the \n depends on the platform? System.lineSeparator() works the same in all platforms! 🙂

src/main/java/BossMan.java
royaliceblue: Agreed! Perhaps it might affect readability as the function names are very similar? Could consider refactoring into a single method!

src/main/java/BossMan.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    private void handleTodoCommand(String commandArgs) {
+        Task todoTask = new Todo(commandArgs);
+        TASK_LIST.addTask(todoTask);
+        echo(todoTask);
+    }
+
+    private void handleDeadlineCommand(String commandArgs) {
+        String[] deadlineArgParts = commandArgs.split("/by", 2);
```
royaliceblue: Is it necessary to add 2 to the split function? Will it be better to declare a constant for it?

src/main/java/Deadline.java
```diff
@@ -0,0 +1,19 @@
+public class Deadline extends Todo {
+    private final String TYPE_SYMBOL = "[D]";
```
royaliceblue: Would it be better not to redeclare the TYPE_SYMBOL as Todo class already has this attribute? 

Perhaps can try to modify this variable in your Deadline constructor than redeclaring?

src/main/java/Event.java
```diff
@@ -0,0 +1,20 @@
+public class Event extends Todo {
+    private final String TYPE_SYMBOL = "[E]";
+    private final String FROM;
+    private final String TO;
+    public Event(String task, String from, String to) {
```
royaliceblue: Would it be clearer to have a linespace between TO variable and Event constructor to enhance clarity?

src/main/java/Main.java
```diff
@@ -0,0 +1,8 @@
+public class Main {
+    public static void main(String[] args) {
+        BossMan chatBot = new BossMan();
```
royaliceblue, line 6: I like the way of abstraction here

royaliceblue (commented):
Very nice way of code abstraction and refactoring.
Just a few nits. Overall great code! 😄

src/main/java/bossman/exceptions/commandexceptions/BossManExceptions.java
```diff
@@ -0,0 +1,5 @@
+package bossman.exceptions.commandexceptions;
+
+public class BossManExceptions extends Exception{
+    //No additional code needed
+}
```
StanleyW00: Try to avoid using empty classes for your exception. Perhaps, add a few methods or a method in handling the exception. Same goes for the other empty exception classes.

src/main/java/bossman/BossMan.java
```diff
@@ -0,0 +1,166 @@
...
+    private String[] parseUserInput(String userInput) {
+        return userInput.trim().split("\\s+", 2);
+    }
+
+    private void handleMarkCommand(String commandArgs) {
+        int number = Integer.parseInt(commandArgs);
+        TASK_LIST.markTask(number);
+    }
+
+    private void handleUnmarkCommand(String commandArgs) throws NumberFormatException{
```
StanleyW00: There is no spacing between the exception and the curly bracket.
`private void handleUnmarkCommand(String commandArgs) throws NumberFormatException{`
Same goes for several methods with this error.

src/main/java/bossman/BossMan.java
```diff
@@ -0,0 +1,166 @@
...
+        String description = eventArgParts[0];
+        String[] eventArgTimeParts = eventArgParts[1].split("/to", 2);
+        if (eventArgTimeParts.length != 2){
+            throw new InvalidEventCommandException();
+        }
+        String from = eventArgTimeParts[0];
+        String to = eventArgTimeParts[1];
+        Task eventTask = new Event(description, from, to);
+        TASK_LIST.addTask(eventTask);
+        echo(eventTask);
```
StanleyW00: Perhaps add blank lines to separate groups of related statements for improving readibility. Same goes for other methods with no blank spaces between several statements.

StanleyW00 (commented):
Overall, good job on maintaining code quality and following the coding standards. There's just a few improvements that could be made.

PR #22, opened 2024-02-03
[JianJiaT] iP

JianJiaT:


src/main/java/UserInterface.java
```diff
@@ -0,0 +1,35 @@
...
+        } else {
+            switch (input) {
+            case "bye":
+                return;
+            case "list":
+                taskManager.listTasks();
+                break;
+            default:
+                taskManager.addTask(input);
+                //Fallthrough
```
webtjs: I like how you added a //Fallthrough comment since it does not have a break statement

src/main/java/UserInterface.java
```diff
@@ -0,0 +1,35 @@
+import java.util.Scanner;
```
webtjs: Explicitly listing the imported classes is good

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,83 @@
+public class TaskManager {
+    public static final int MAX_TASKS = 100;
+    public static final int TODO_TASK_NAME_BEGIN_INDEX = 5;
+    public static final int DEADLINE_TASK_NAME_BEGIN_INDEX = 9;
+    public static final int SKIP_BY_IN_DEADLINE = 4;
+    public static final int EVENT_TASK_NAME_BEGIN_INDEX = 6;
+    public static final int SKIP_FROM_IN_EVENT = 6;
+    public static final int SKIP_TO_IN_EVENT = 4;
+    private int currIndex;
+    private Task[] tasks;
```
webtjs: Appropriate use of plural form to indicate multiple tasks

src/main/java/Task.java
```diff
@@ -0,0 +1,38 @@
+public class Task {
+    private String name;
+    private boolean isDone;
```
webtjs: Suitable name, sounds like a boolean variable

src/main/java/Task.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public String getDoneStatus() {
+        if (isDone) {
+            return "X";
+        }
```
webtjs: Single statement conditional wrapped in curly braces is good

webtjs (commented):
Nice work, overall the code follows the coding standard (layout, statements, and naming of variables, classes, & methods).

src/main/java/Dor.java
```diff
@@ -0,0 +1,10 @@
+public class Dor {
+    public static void main(String[] args) {
+
+        System.out.println("Hello! I'm Dor");
+        System.out.println("What can I do for you?");
+        UserInterface UI = new UserInterface();
+        UI.processInput();
+        System.out.println("Bye. Hope to see you again soon!");
+    }
+}
```
hwc0419: Perhaps you can improve this by following the Single Layer of Abstraction Principle (SLAP), as the assignment statement in line 6 is on a different layer of abstraction than the other statements. You can create a method within the same class and place lines 6 and 7 inside. For example : 
```suggestion
public class Dor {
    public static void main(String[] args) {

        System.out.println("Hello! I'm Dor");
        System.out.println("What can I do for you?");
        startListening();
        System.out.println("Bye. Hope to see you again soon!");
    }
    public void startListening() {
        UserInterface UI = new UserInterface();
        UI.processInput();
    }
}
```

src/main/java/Task.java
```diff
@@ -0,0 +1,38 @@
+public class Task {
+    private String name;
+    private boolean isDone;
+
+    public Task() {
+        this.name = "placeholder";
+        this.isDone = false;
+    }
```
hwc0419: You can consider using the keyword "this" to shorten the code and improve readability. For example : 
```suggestion
    public Task() {
        this("placeholder");
    }
```

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,83 @@
+public class TaskManager {
+    public static final int MAX_TASKS = 100;
+    public static final int TODO_TASK_NAME_BEGIN_INDEX = 5;
+    public static final int DEADLINE_TASK_NAME_BEGIN_INDEX = 9;
+    public static final int SKIP_BY_IN_DEADLINE = 4;
+    public static final int EVENT_TASK_NAME_BEGIN_INDEX = 6;
+    public static final int SKIP_FROM_IN_EVENT = 6;
+    public static final int SKIP_TO_IN_EVENT = 4;
```
hwc0419: I'm not sure if I like these constant names because the names do not give a clear idea of what they are used for.

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,83 @@
...
+        if (taskToAdd.contains("todo")) {
+            processToDo(taskToAdd);
+        } else if (taskToAdd.contains("deadline")) {
+            processDeadline(taskToAdd);
+        } else if (taskToAdd.contains("event")) {
+            processEvent(taskToAdd);
+        } else {
+            System.out.println("ERROR: Invalid input");
+            return;
+        }
```
hwc0419: Any reason why the contains method is used to check for the type of command?
Consider the input "deadline todo homework".  Will this task be added as a deadline task or todo task? Was the code intended to work such that the todo tasks has a higher precedence over deadline tasks?

src/main/java/UserInterface.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public void processInput() {
+        String input = reader.nextLine();
+        if (input.contains("unmark")) {
+            int taskIndex = Character.getNumericValue(input.charAt(INDEX_MARK_NOT_DONE)) - 1;
+            taskManager.markTask(taskIndex, false);
+        } else if (input.contains("mark")) {
+            int taskIndex = Character.getNumericValue(input.charAt(INDEX_MARK_DONE)) - 1;
+            taskManager.markTask(taskIndex, true);
+        } else {
```
hwc0419: I like how you avoided deep nesting using else if statements

src/main/java/UserInterface.java
```diff
@@ -0,0 +1,35 @@
...
+                return;
+            case "list":
+                taskManager.listTasks();
+                break;
+            default:
+                taskManager.addTask(input);
+                //Fallthrough
+            }
+        }
+        processInput();
```
hwc0419: Any reason why you chose to do a recursive call instead of a while loop?

hwc0419 (commented):
Overall,  the code was well written. I liked the fact that you did UI testing as well. There are some area of improvements which can be worked on to make the code better.

src/main/java/Dor.java
JianJiaT: Implemented, thanks for the suggestion

src/main/java/Task.java
JianJiaT: Implemented, thanks for the suggestion

src/main/java/TaskManager.java
JianJiaT: Changed constant names to be clearer

src/main/java/TaskManager.java
JianJiaT: I neglected to consider the cases where (1) there are more than one keyword in the input and (2) the keyword is not the first word in the input. Changed the if conditions to check for these

src/main/java/UserInterface.java
JianJiaT: Can't seem to recall a reason for the recursion 🤔 Changed to while loop since loops are more efficient (to my knowledge?)

hwc0419: With the previous recursive call, the call stack may overflow if the user enters too many commands (about 100000) during program execution, and you can get a stack overflow error, causing the program to crash

src/main/java/logic/TaskManager.java
```diff
@@ -0,0 +1,140 @@
...
+import exceptions.EventToBeforeFromException;
+import exceptions.MarkUnmarkIndexOutOfBoundsException;
+import tasks.Task;
+import tasks.ToDo;
+import tasks.Deadline;
+import tasks.Event;
+
+public class TaskManager {
+    public static final int MAX_TASKS = 100;
+    public static final int TODO_LENGTH= 5;
```
Jonoans: Maybe a missing space here

src/main/java/logic/TaskManager.java
```diff
@@ -0,0 +1,140 @@
...
+        this.currIndex = 0;
+        this.tasks = new Task[MAX_TASKS];
+    }
+
+    public void addTask(String taskToAdd) throws Exception {
+        String[] taskAsArray = taskToAdd.split(" ");
+        if (taskAsArray.length == 1) {
+            throw new TaskNoNameException();
+        }
+        String taskType = taskAsArray[0];
```
Jonoans: Try to avoid magic numbers

src/main/java/logic/TaskManager.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    private void processEvent(String taskToAdd) throws Exception {
+        if (!(taskToAdd.contains("/from "))) {
+            throw new EventNoFromDateTimeException();
+        }
+        if (!(taskToAdd.contains("/to "))) {
+            throw new EventNoToDateTimeException();
+        }
+        if (taskToAdd.indexOf("/to ") < taskToAdd.indexOf("/from ")) {
```
Jonoans: Magic strings like these could be move to a class-level constant variable, especially since they are used more than once as well

Jonoans (commented):
Overall, good job with the code. Just some minor issues here and there, otherwise looks well done.

PR #25, opened 2024-02-03
[Yeo Zhi Shen] ip

yeozhishen:


src/main/java/Parser.java
```diff
@@ -0,0 +1,21 @@
+import java.util.Scanner;
+public class Parser {
```
Zhengwinter: Clever use of abstraction to reduce repetition in code and improve modularization.

src/main/java/Fido.java
```diff
@@ -31,15 +31,44 @@ private String processInputCommand(String command){
             return handleTaskMarking();
         case UNMARK:
             return handleUnmarkingTask();
+        case TODO:
+            return addTodo();
+        case DEADLINE:
+            return addDeadline();
+        case EVENT:
+            return addEvent();
```
Zhengwinter: Clear and descriptive naming of helper methods

src/main/java/ParserRegex.java
```diff
@@ -1,14 +1,14 @@
-public enum PARSER_REGEX {
+public enum ParserRegex {
```
Zhengwinter: Good change! Classes should be PascalCase!

src/main/java/Fido.java
```diff
@@ -31,15 +31,44 @@ private String processInputCommand(String command){
...
         }
     }
     private void Exit() {
         userInterface.printExitMessage();
         System.exit(0);
     }
-
+    private String addTodo() throws IllegalArgumentException {
+        if(!inputParser.isValidTodo()) {
+            throw new IllegalArgumentException("invalid Todo");
```
Zhengwinter: Apologies I may not fully understand how the throw statement works here but how do you classify an invalid statement as a invalid ToDo /Event/Deadline?

src/main/java/Event.java
```diff
@@ -0,0 +1,15 @@
+public class Event extends Task {
+    protected String startTime;
+    protected String endTime;
+    public Event(String task, String startTime, String endTime) {
+        super(task, 'E');
+        this.startTime = startTime;
+        this.endTime = endTime;
+    }
+    @Override
+    public String toString() {
```
Zhengwinter: Could consider a more descriptive name instead of toString, perhaps like outputString

src/main/java/Event.java
```diff
@@ -0,0 +1,15 @@
...
+        return super.toString()
+                + " (" + "from: " + startTime
+                + "to: " + endTime + ")";
+    }
```
Zhengwinter: An alternative way of structuring your Event/Deadline classes that you could consider is adding a class for handling all outputs. Then you could use some setter and getter methods for the fields of these Task subclasses. It may be cleaner and more modularized that way

src/main/java/Parser.java
```diff
@@ -28,6 +28,60 @@ public String getUserInputString() {
...
+        return byKeywordIndex != -1 && !atEndsOfCommandArgumentList(byKeywordIndex);
+    }
+    public boolean isValidEvent() {
+        int fromKeywordIndex = indexOfKeywordInCommandArguments(PARSER_REGEX.FROM.string);
+        int toKeywordIndex = indexOfKeywordInCommandArguments(PARSER_REGEX.TO.string);
+        int lastElementIndex = inputCommandArguments.length - 1;
+        return fromKeywordIndex != -1 && toKeywordIndex != -1
+                && !atEndsOfCommandArgumentList(fromKeywordIndex)
+                && !atEndsOfCommandArgumentList(toKeywordIndex)
+                && !consecutiveInArgumentList(fromKeywordIndex, toKeywordIndex);
```
Zhengwinter: Could improve readability with a switch case and a flag variable

Zhengwinter (commented):
Generally well written with clear applications of abstraction, inheritance concepts. Some minor changes have been suggested, I think the main ones include modifying the Event/Deadline classes as well as some possible readability fixes. Keep up the good work!

src/main/java/Fido.java
yeozhishen: I defined an invalid todo when the user inputs the todo command without specifying any following tasks.
An invalid event/deadline is when the user does not specify BY or TO/FROM.

src/main/java/Event.java
yeozhishen: I deliberately chose to overrise the toString method so that in the even that there is a need to print the tasks/event/deadline using the System.out.print, the toString method will be invoked

Zhengwinter: Ok, that makes sense. So it accounts for the cases when you print those objects (toDo/Event/Deadline) directly, so it's more comprehensive in a way?

src/main/java/Fido.java
Zhengwinter: Alright, so to clarify, you are accounting for the situations where the correct command words were given but the input was structured in an invalid way, which falls into the invalid toDo/Event/Deadline categories depending on the command. 
Have you also handled inputs without any command words?

yeozhishen: Thanks for pointing it out! I have yet to account for empty inputs as I did think of that option yet. Going to handle it as part of this week's level 5 :)

shawnpong (commented):
I really like that you separated all your methods to improve legibility! Methods used also have good error handling (great work implementing try catch blocks).

src/main/java/Fido.java
```diff
@@ -0,0 +1,99 @@
...
+        case UNMARK:
+            return handleUnmarkingTask();
+        case TODO:
+            return addTodo();
+        case DEADLINE:
+            return addDeadline();
+        case EVENT:
+            return addEvent();
+        default:
+            return Commands.INVALID_COMMAND.string;
```
shawnpong: Switch case usage makes it very readable, good job!

src/main/java/Parser.java
```diff
@@ -0,0 +1,138 @@
...
+        ParserRegex[] keywordsToLookOutFor = {ParserRegex.BY};
+        if(inputCommandArguments == null || !onlyContainsKeywords(keywordsToLookOutFor)){
+            return false;
+        }
+        int byKeywordIndex = indexOfKeywordInCommandArguments(ParserRegex.BY.string);
+        return byKeywordIndex != DOES_NOT_EXIST
+                && !atEndsOfCommandArgumentList(byKeywordIndex);
+    }
+    public boolean isValidEvent() {
+        ParserRegex[] keywordsToLookOutFor = {ParserRegex.TO, ParserRegex.FROM};
```
shawnpong: Great idea, making sure to check for validity of the input/event!

src/main/java/Commands.java
```diff
@@ -0,0 +1,22 @@
+public enum Commands {
+    LIST("list"),
+    TODO("todo"),
+    DEADLINE("deadline"),
+    EVENT("event"),
+    EXIT("bye"),
+    MARK("mark"),
+    UNMARK("unmark"),
+    INVALID_COMMAND("invalid command!");
+    public final String string;
```
shawnpong: Good way to define all the commands, to make editing your code in the future easy and more recognizable!

src/main/java/ErrorMessages.java
```diff
@@ -0,0 +1,14 @@
...
+    INVALID_TODO("Invalid todo. Input todo in the form of: todo [task]"),
+
+    MISSING_INPUT("No input detected, enter an input command"),
+    FILE_ERROR("error regarding file IO"),
+    INDEX_OUT_OF_BOUNDS("Index given is out of bounds");
+    public final String string;
+    private ErrorMessages(String string) {
+        this.string = string;
+    }
+}
```
Jonoans: Nice use of enums!

src/main/java/Commands.java
```diff
@@ -0,0 +1,23 @@
...
+    EXIT("bye"),
+    MARK("mark"),
+    UNMARK("unmark"),
+    DELETE("delete"),
+    INVALID_COMMAND("invalid command!");
+    public final String string;
+    private Commands(String string) {
+        this.string = string;
+    }
+    public static Commands getCommandEnumeration (String command) {
```
Jonoans: Extra space here

src/main/java/Parser.java
```diff
@@ -0,0 +1,138 @@
...
+        inputString = null;
+        inputCommand = null;
+        inputCommandArguments = null;
+    }
+    public void collectUserInput() throws FidoException {
+        clearAllInputs();
+        inputString = stdin.nextLine();
+        String[] individualWords = inputString.trim().split(" ");
+        int last_word_index = individualWords.length;
+        if(individualWords.length == 0) {
```
Jonoans: Missing space here

src/main/java/Task.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+    public char getType() {
+        return type;
+    }
+    public Task(String task, char type) {
+        this.task = task;
+        this.type = type;
+        setNotDone();
+    }
+   public String toString() {
```
Jonoans: Indentation issue here

Jonoans (commented):
Good job, you have done great in terms of avoid magic strings and numbers. Just be a little mindful of the issues with missing/extra spaces in your code.

PR #26, opened 2024-02-03
[Daryl Tay Chin Kian] iP

daryltay415:


src/main/java/Duke.java
```diff
@@ -1,10 +1,84 @@
+import java.util.Scanner;
 public class Duke {
     public static void main(String[] args) {
```
cheeseong2001: I noticed your entire specific code is in `main()`. Perhaps you can extract into functions and abstract the logic?

src/main/java/Duke.java
```diff
@@ -1,10 +1,84 @@
...
+            } else if(line.startsWith("deadline")){
+                // Adds a deadline task into the list
+                String []sentence = line.split("deadline|/by");
+                Task newTask = new Deadlines(sentence[1], sentence[2]);
+                taskList.addTask(newTask);
+                System.out.println("    ____________________________________________________________");
+                System.out.println("     Got it. I've added this task:");
+                System.out.println("       " + newTask);
+                System.out.println("     Now you have " + taskList.getNoOfTasks() + " tasks in the list.");
+                System.out.println("    ____________________________________________________________");
```
cheeseong2001: These lines to print are repeated throughout the conditional structure, perhaps you can use inheritance to perform this?

src/main/java/Task.java
```diff
@@ -0,0 +1,24 @@
+public class Task {
+   /** Task description */
+    private String task;
+    /** Task status */
+   private boolean taskIsDone = false;
+   public Task(String line){
```
cheeseong2001: There are inconsistency in the number of spaces here. Please follow the coding standard.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,62 @@
...
+        noOfTasks++;
+    }
+
+    /**
+     * Checks the task in the task list
+     * @param taskNumber The index of the task
+     */
+    public void checkTask(int taskNumber){
+        tasks[taskNumber-1].setTaskStatus(true);
+        System.out.println("      " + tasks[taskNumber-1]);
```
cheeseong2001: Perhaps you can use other printing characters for the spaces? I suggest `\t` char for tabs

src/main/java/TaskList.java
```diff
@@ -0,0 +1,62 @@
+public class TaskList {
+    /** Array of tasks */
+    private Task[] tasks;
+
+    /** Number of tasks */
+    private int noOfTasks = 0;
+    public TaskList(int taskLength) {
+        tasks = new Task[taskLength];
+    }
```
cheeseong2001: Unnecessary step of reassigning variable. Assume that the `taskLength` is fixed at 100 as suggested in the assignment, then you can immediately declare and assign into the variable in line 3

src/main/java/Duke.java
```diff
@@ -1,10 +1,84 @@
...
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        // Creates a class of TaskList
+        TaskList taskList = new TaskList(100);
+        String line;
```
vibes-863: The variable name is a bit misleading. Do you think it would be clearer if you renamed it to something like "inputLine"?

src/main/java/Duke.java
```diff
@@ -1,10 +1,84 @@
...
+        Scanner in = new Scanner(System.in);
+        boolean userSaidBye = false;
```
vibes-863: I think maybe you could rename the boolean  variable to something like "hasEnteredBye" ? I think it would help make the code more readable.

src/main/java/Task.java
```diff
@@ -0,0 +1,24 @@
+public class Task {
+   /** Task description */
+    private String task;
```
vibes-863: The indentation has 5 spaces instead of 4. It was probably by accident. I would recommend adding an auto formatter on save which would help fix these accidents. You can search "auto formatter on save in Intellij IDEA" on Google.

src/main/java/Duke.java
```diff
@@ -1,10 +1,84 @@
...
+        // Welcomes user
+        System.out.println("    ____________________________________________________________");
+        System.out.println("    Hello! I'm PoratoBot");
+        System.out.println("    How can I assist you mortal?");
+        System.out.println("    ____________________________________________________________");
+        while (!userSaidBye) {
+            // Takes in user input
+            line = in.nextLine();
+            if (line.equals("bye")) {
```
vibes-863: Good job in maintaining the layout for the if-else statements. Usually people do mess up when there are so many if-else statements, but you haven't, which is great!

vibes-863 (commented):
Overall, the code looks good. Just a few suggestions have been made which would enhance your code readability even further.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,228 @@
...
+        System.out.println("    How can I assist you mortal?");
+        printLine();
+    }
+
+    /**
+     * Loads the file contents into the task list
+     * @param tasklist The task list
+     * @throws FileNotFoundException File is not found exception
+     * @throws DukeException File is corrupted exception
+     */
```
aditichadha1310: Good work with the comments. Increases the code readability.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,228 @@
...
+                Task deadlineTask = new Deadlines(sentence[2], sentence[3]);
+                tasklist.addTask(deadlineTask);
+                if(sentence[1].equals(" 1 ")){
+                    deadlineTask.setTaskStatus(true);
+                }
+                break;
+            case "E":
+                Task eventTask = new Events(sentence[2], sentence[3], sentence[4]);
+                tasklist.addTask(eventTask);
+                if(sentence[1].equals(" 1 ")){
```
aditichadha1310: Avoid usage of Magic strings.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,228 @@
...
+     */
+    public static boolean isNumeric(String str) {
+        try {
+            Double.parseDouble(str);
+            return true;
+        } catch(NumberFormatException e){
+            return false;
+        }
+    }
+    public static void main(String[] args) {
```
aditichadha1310: Avoid long methods. Take corrective action if method exceeds 30 lines of code.

aditichadha1310 (commented):
Overall, good job!

PR #31, opened 2024-02-03
[NicholasTanYY] iP

NicholasTanYY:


src/main/java/BobBot.java
```diff
@@ -0,0 +1,93 @@
+import java.util.Scanner;
+
+public class BobBot {
+    
+    private static Task[] allTasks = new Task[100];
```
kyhjonathan: Maybe can replace 100 as a static final int declaration at the top as MAX_INT? This is to avoid "magic numbers"
```suggestion
    private static Task[] allTasks = new Task[MAX_INT];
```

src/main/java/BobBot.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+    public static void main(String[] args) {
+        greet();
+
+        String line;
+        Scanner in = new Scanner(System.in);
+
+        line = in.nextLine();
+
+        while (!line.equalsIgnoreCase("bye")) {
```
kyhjonathan: I like how methods were implemented instead of writing the entire logic in the else-if blocks!

kyhjonathan (commented):
Looks good. 👍 
Just some minor styling

src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
...
+    public void markAsDone() {
+        this.isDone = true;
+    }
+    
+    public void markAsUndone() {
+        this.isDone = false;
+    }
+    
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); // mark done task with X
```
Elyovs: nice work on following coding standards for trailing comment

src/main/java/Event.java
```diff
@@ -0,0 +1,28 @@
+public class Event extends Task {
+
+    protected String task;
+    protected String from;
+    protected String to;
+
+    public Event(String description) {
+        super(description);
+        this.task = this.description.substring(
+                "event".length(),
```
Elyovs: it's great that you wrapped lines with 8 spaces indentation

src/main/java/BobBot.java
```diff
@@ -0,0 +1,104 @@
...
+
+            if (line.equalsIgnoreCase("list")) {
+                listItems();
+            } else if (line.startsWith("mark")) {
+                mark(line);
```
Elyovs: correct layout using Egyptian style brackets and 4 spaces indentation

src/main/java/BobBot.java
```diff
@@ -0,0 +1,104 @@
...
+        int taskNumToUnmark = Integer.parseInt(line.substring(6).trim()) - 1;
+        allTasks[taskNumToUnmark].markAsUndone();
+        drawLine(true);
+        System.out.println("\tAlright! Unmarking this task:");
+        System.out.printf("\t[%s] %s\n", allTasks[taskNumToUnmark].getStatusIcon(),
+                allTasks[taskNumToUnmark].getDescription());
+        drawLine(true);
+    }
+
+    private static void listItems() {
```
Elyovs: good work on methods names using camel case

src/main/java/BobBot.java
```diff
@@ -0,0 +1,104 @@
...
+        allTasks[taskNumToMark].markAsDone();
+        drawLine(true);
+        System.out.println("\tGot it! Marking this task as done:");
+        System.out.printf("\t[%s] %s\n", allTasks[taskNumToMark].getStatusIcon(),
+                allTasks[taskNumToMark].getDescription());
+        drawLine(true);
+    }
+
+    private static void unmark(String line) {
+        int taskNumToUnmark = Integer.parseInt(line.substring(6).trim()) - 1;
```
BryanCastorius: I believe the variable naming `taskNumToUnmark` could be better as Num is a bit ambiguous. I would suggest naming it to  `taskNumberToUnmark`.

src/main/java/BobBot.java
```diff
@@ -0,0 +1,104 @@
...
+    private static int numTasks = 0;
+
+    private static void mark(String line) {
+        int taskNumToMark = Integer.parseInt(line.substring(4).trim()) - 1;
+        allTasks[taskNumToMark].markAsDone();
+        drawLine(true);
+        System.out.println("\tGot it! Marking this task as done:");
+        System.out.printf("\t[%s] %s\n", allTasks[taskNumToMark].getStatusIcon(),
+                allTasks[taskNumToMark].getDescription());
+        drawLine(true);
```
BryanCastorius: I believe you can extract these printing lines to a method like `PrintMessageMarkAsDone() ` to improve code readability and enforce SLAP.

The code mentioned:
```
drawLine(true);
System.out.println("\tGot it! Marking this task as done:");
System.out.printf("\t[%s] %s\n", allTasks[taskNumToMark].getStatusIcon(), allTasks[taskNumToMark].getDescription());
drawLine(true);
```

src/main/java/BobBot.java
```diff
@@ -0,0 +1,104 @@
...
+
+        Task newTask = null;
+        if (line.startsWith("todo")) {
+            newTask = new Todo(line);
+        } else if (line.startsWith("deadline")) {
+            newTask = new Deadline(line);
+        } else if (line.startsWith("event")) {
+            newTask = new Event(line);
+        }
+
```
BryanCastorius: There are no default "else" case for the if-else-if statements.

src/main/java/BobBot.java
BryanCastorius: The method naming `listItems` is a bit ambiguous, I believe it's better to rename it to something like `printListItems`, because we are basically printing all the items.

BryanCastorius (commented):
Overall, a good coding standard has been followed, however there are several variable naming confusions found.

src/main/java/BobBot.java
NicholasTanYY: Good idea! I will take that into account

src/main/java/BobBot.java
NicholasTanYY: Thanks for pointing out!

src/main/java/BobBot.java
NicholasTanYY: Great, I like that approach as well

src/main/java/Event.java
NicholasTanYY: Thanks!

src/main/java/BobBot.java
NicholasTanYY: Got it, I will implement that. Thanks!

src/main/java/BobBot/BobBot.java
```diff
@@ -0,0 +1,323 @@
...
+                case UNMARK:
+                    task.markAsUndone();
+                    printTaskOperationMessage(task, "Unmarking this task:");
+                    break;
+                case DELETE:
+                    allTasks.remove(taskNumber);
+                    printTaskOperationMessage(task, "Deleting this task:");
+                    numTasks -= 1;
+                    break;
+            }
```
Musfirahe0556596: the switch statement should have the following form:

```
switch (condition) {
case ABC:
    statements;
    // Fallthrough
case DEF:
    statements;
    break;
case XYZ:
    statements;
    break;
default:
    statements;
    break;
}
```

src/main/java/BobBot/BobBot.java
```diff
@@ -0,0 +1,323 @@
...
+import BobBot.exceptions.InvalidTodoException;
+import BobBot.tasks.Deadline;
+import BobBot.tasks.Event;
+import BobBot.tasks.Task;
+import BobBot.tasks.Todo;
+
+public class BobBot {
+
+    private static ArrayList<Task> allTasks = new ArrayList<>();
+    private static int numTasks = 0;
```
Musfirahe0556596: nothing major but perhaps this could be renamed better for better readability :] (e.g. numberOfTasks)

src/main/java/BobBot/BobBot.java
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    private static void printTaskOperationMessage(Task task, String operationMessage) {
+        drawLine(true);
+        System.out.println("\tGot it! " + operationMessage);
+        System.out.println("\t  " + task.toString());
+        drawLine(true);
+    }
+
+    private static void printErrorMessage(int taskNumber) {
```
Musfirahe0556596: for this method, perhaps for better readability you could indicate what kind of error message! e.g. printNonExistentTaskErrorMessage or something appropriate 👍

Musfirahe0556596 (commented):
Nice! I could barely find anything but some minor coding standard/quality to consider :)

src/main/java/BobBot/BobBot.java
NicholasTanYY: Must have slipped my mind!

src/main/java/BobBot/BobBot.java
NicholasTanYY: I will implement that, thanks for pointing out :)

src/main/java/BobBot/BobBot.java
NicholasTanYY: ok!

PR #46, opened 2024-02-04
[jasraa] ip

jasraa:


src/main/java/Joey.java
```diff
@@ -0,0 +1,218 @@
...
+
+        Scanner scanner = new Scanner(System.in);
+        Task[] tasks = new Task[100];
+        int taskCount = 0;
+
+        while (true) {
+            String userCommand = scanner.nextLine();
+            System.out.println("____________________________________________________________");
+
+            if (userCommand.equalsIgnoreCase("bye")) {
```
wallywallywally: I like that you ignore the case!

src/main/java/Joey.java
```diff
@@ -0,0 +1,218 @@
...
+                break;
+            } else if (userCommand.equalsIgnoreCase("list")) {
+                System.out.println(" Here are the tasks in your list:");
+                for (int i = 0; i < taskCount; i++) {
+                    Task task = tasks[i];
+                    System.out.println(" " + (i + 1) + "." + task.getType() + "[" + (task.isDone() ? "X" : " ") + "] " + task.getDescription());
+                }
+                System.out.println("____________________________________________________________");
+            } else if (userCommand.startsWith("todo")) {
+                String description = userCommand.substring(5);
```
wallywallywally: Should 5 be turned into a variable to remove the magic number?
```suggestion
                int todoLength = "todo".length();
                String description = userCommand.substring(todoLength);
```

src/main/java/Joey.java
```diff
@@ -0,0 +1,218 @@
...
+
+    public void markNotDone() {
+        isDone = false;
+    }
+    public String getStatusIcon() {
+        return (isDone ? "[X]" : "[ ]");
+    }
+
+    public String getType() {
+        return "";
```
wallywallywally: Should this function be removed as it currently does not return anything?

src/main/java/Joey.java
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    public String toString() {
+        return getType() + " " + getStatusIcon() + " " + description;
+    }
+}
+
+//class Todo extends Task {
```
wallywallywally: Should this part be removed as it is not being used?

src/main/java/Joey.java
```diff
@@ -0,0 +1,218 @@
...
+        super(description);
+    }
+
+    public String getType() {
+        return "[T]";
+    }
+}
+
+
+//class Deadline extends Task {
```
wallywallywally: Should this code be removed as it is not being used?

src/main/java/Joey.java
```diff
@@ -0,0 +1,218 @@
...
+        this.by = by;
+    }
+
+    public String getType() {
+        return "[D]";
+    }
+}
+
+
+//class Event extends Task {
```
wallywallywally: Should this code be removed as it is not being used?

wallywallywally (commented):
Overall, LGTM! It is clean and organised well, however there are some parts that can be removed, like the parts that are commented are implemented right below!

src/main/java/Joey.java
```diff
@@ -0,0 +1,218 @@
+import java.util.Scanner;
+
+class Task {
```
nkotaa: Should this be extracted out into its own class file? Likewise with all other non-main classes.

src/main/java/Joey.java
```diff
@@ -0,0 +1,218 @@
...
+                System.out.println("   " + tasks[taskCount - 1]);
+                System.out.println(" Now you have " + taskCount + " tasks in the list.");
+                System.out.println("____________________________________________________________");
+            } else if (userCommand.startsWith("mark")) {
+                int taskIndex = Integer.parseInt(userCommand.split(" ")[1]) - 1;
+                if (taskIndex >= 0 && taskIndex < taskCount) {
+                    Task task = tasks[taskIndex];
+                    task.markDone();
+                    System.out.println(" Nice! I've marked this task as done:");
+                    System.out.println("   [" + (task.isDone() ? "X" : " ") + "] " + task.getDescription());
```
nkotaa:
```suggestion
                    System.out.println(task);
```
Suffices to print all the information needed since toString() has been appropriately overridden. Likewise for all subsequent occurrences.

src/main/java/Joey.java
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    public String toString() {
+        return super.toString() + " (from: " + from + " to: " + to + ")";
+    }
+}
+
+
+public class Joey {
+    public static void main(String[] args) {
```
nkotaa, line 100: The main method could be much more readable if it were broken down into smaller sub-functions. One such function could be called within each if-condition, given that they share much of the same logic.

src/main/java/Joey.java
```diff
@@ -0,0 +1,218 @@
...
+                taskCount++;
+                System.out.println(" Got it. I've added this task:");
+                System.out.println("   " + tasks[taskCount - 1]);
+                System.out.println(" Now you have " + taskCount + " tasks in the list.");
+                System.out.println("____________________________________________________________");
+            } else if (userCommand.startsWith("event")) {
+                String[] parts = userCommand.substring(6).split("/from|/to");
+                String description = parts[0].trim();
+                String from = parts[1].trim();
+                String to = parts[2].trim();
```
nkotaa: Nice, concise way of parsing string.

src/main/java/Joey.java
```diff
@@ -0,0 +1,218 @@
...
+                if (taskIndex >= 0 && taskIndex < taskCount) {
+                    Task task = tasks[taskIndex];
+                    task.markDone();
+                    System.out.println(" Nice! I've marked this task as done:");
+                    System.out.println("   [" + (task.isDone() ? "X" : " ") + "] " + task.getDescription());
+                    System.out.println("____________________________________________________________");
+                } else {
+                    System.out.println(" Task not found. Please enter a valid task number OR enter 'list' to view your current list:)");
+                    System.out.println("____________________________________________________________");
+                }
```
nkotaa: Rather than checking for the validity of taskIndex, break out of the branch prematurely if it is invalid. Then the "happy path" will be more obvious with one less level of indentation.

nkotaa (commented):
Overall could benefit from decomposing into smaller classes and functions.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,12 @@
+class Deadline extends Task {
```
aditichadha1310: Comments are missing. Please add them for better code readability.

src/main/java/Joey.java
```diff
@@ -0,0 +1,137 @@
+import java.util.Scanner;
+
+public class Joey {
+    public static void main(String[] args) throws JoeyException {
```
aditichadha1310: This is a very long method. Try to avoid long methods. Take corrective action if method exceeds 30 lines of code.

src/main/java/Joey.java
```diff
@@ -0,0 +1,137 @@
...
+                    System.out.println("OOPS!!! " + e.getMessage());
+                    System.out.println("____________________________________________________________");
+                }
+            } else if (userCommand.startsWith("deadline")) {
+                try {
+                    String[] parts = userCommand.substring(9).split("/by");
+                    if (parts.length != 2) {
+                        throw new JoeyException("Hey girlie, please provide both description and deadline for the deadline task. Follow this format instead: deadline return book /by Sunday");
+                    }
+
```
aditichadha1310: Avoid the arrowhead style code as much a possible.

src/main/java/Joey.java
```diff
@@ -0,0 +1,137 @@
...
+                    tasks[taskCount] = new Deadline(description, by);
+                    taskCount++;
+                    System.out.println(" Got it. I've added this task:");
+                    System.out.println("   " + tasks[taskCount - 1]);
+                    System.out.println(" Now you have " + taskCount + " tasks in the list.");
+                    System.out.println("____________________________________________________________");
+                } catch (StringIndexOutOfBoundsException e) {
+                    System.out.println("OOPS!!! Hey girlie, please provide both description and deadline for the deadline task. Follow this format instead: deadline return book /by Sunday");
+                    System.out.println("____________________________________________________________");
```
aditichadha1310: You might want to save this horizontal line in a variable and reference that instead of using this everywhere.

src/main/java/Joey.java
```diff
@@ -0,0 +1,137 @@
...
+                try {
+                    String[] parts = userCommand.split("/from");
+                    if (parts.length != 2) {
+                        throw new JoeyException("Hey girlie, please provide both description, start time, and end time for the event task. Follow this format instead: event project meeting /from Mon 2pm /to 4pm");
+                    }
+
+                    String description = parts[0].substring(6).trim();
+                    String[] timingParts = parts[1].split("/to");
+                    if (timingParts.length != 2) {
+                        throw new JoeyException("Hey girlie, please provide both description, start time, and end time for the event task. Follow this format instead: event project meeting /from Mon 2pm /to 4pm");
```
aditichadha1310: This code line exceeds the number of characters allowed per line. Please read the guidelines and follow.

PR #47, opened 2024-02-04
[Li Kunjun] iP

StargazerX0:
add several function to chatbot...


src/main/java/Conversation.java
```diff
@@ -0,0 +1,113 @@
...
+        Scanner userInput = new Scanner(System.in);
+
+        while (userInput.hasNextLine()) {
+            String receivedMessage = userInput.nextLine().trim();
+            String[] commandParts = receivedMessage.split(" ", 2);
+            String receivedCommand = commandParts[0];
+            String details = commandParts.length > 1 ? commandParts[1] : "";
+
+            switch (receivedCommand) {
+            case "list":
```
ChenKangg: code follows coding standard

src/main/java/Conversation.java
```diff
@@ -0,0 +1,113 @@
...
+            case "mark":
+                int taskIndexMark = Integer.parseInt(details) - 1;
+
+                if (taskList.isCountValid(taskIndexMark)) {
+                    System.out.print(HORIZONTAL_LINE);
+                    taskList.markTask(taskIndexMark);
+                    System.out.print(HORIZONTAL_LINE);
+                }
+                else {
+                    System.out.print(HORIZONTAL_LINE);
```
LowTL: Perhaps you could abstract all the `System.out.print(HORIZONTAL_LINE);` into the top and bottom of the switch-case, rather than in every case.

src/main/java/Task.java
```diff
@@ -0,0 +1,24 @@
...
+
+    public void mark() {
+        this.isDone = true;
+    }
+
+    public void unmark() {
+        this.isDone = false;
+    }
+
+    public String toString() {
```
LowTL: You forgot to add `@Override` for the overriding of the `toString()` method

src/main/java/Conversation.java
```diff
@@ -0,0 +1,113 @@
...
+                int taskIndexUnmark = Integer.parseInt(details) - 1;
+
+                if (taskList.isCountValid(taskIndexUnmark)) {
+                    System.out.print(HORIZONTAL_LINE);
+                    taskList.unmarkTask(taskIndexUnmark);
+                    System.out.print(HORIZONTAL_LINE);
+                }
+                else {
+                    System.out.print(HORIZONTAL_LINE);
+                    System.out.println("invalid index");
```
LowTL: Since there is repetition in the code, perhaps you could abstract this part into the Task class?

src/main/java/Conversation.java
```diff
@@ -0,0 +1,113 @@
...
+                    System.out.print(HORIZONTAL_LINE);
+                }
+                break;
+
+            case "unmark":
+                int taskIndexUnmark = Integer.parseInt(details) - 1;
+
+                if (taskList.isCountValid(taskIndexUnmark)) {
+                    System.out.print(HORIZONTAL_LINE);
+                    taskList.unmarkTask(taskIndexUnmark);
```
LowTL: This works, but consider using a setter instead of 2 separate functions, something like

```
public void setMarked(boolean done) {
    this.isDone = done;
}
```

LowTL (commented):
Well done so far!

src/main/java/Conversation.java
StargazerX0: yeah, thanks for your suggestions

src/main/java/alpaca/Alpaca.java
```diff
@@ -0,0 +1,20 @@
+package alpaca;
```
aditichadha1310, line 1: Good job in creating a package.

src/main/java/alpaca/Alpaca.java
```diff
@@ -0,0 +1,20 @@
...
+
```
aditichadha1310, line 2: Comments are missing. Good to include them for better readability.

src/main/java/alpaca/UI/InputManager.java
```diff
@@ -0,0 +1,61 @@
...
+import alpaca.exceptions.InvalidCommandException;
+
+public class InputManager {
+    private Scanner scanner;
+
+    public InputManager() {
+        this.scanner = new Scanner(System.in);
+    }
+
+    public void listenForInput(LogicManager logicManager){
```
aditichadha1310: Avoid long methods. Take corrective action when it exceeds 30 lines of code.

aditichadha1310 (commented):
Overall, very good job! Keep it up!

PR #48, opened 2024-02-04
[Xb990219] iP

Xb990219:


src/main/java/ChatBBT.java
```diff
@@ -1,23 +1,34 @@
 import java.util.Scanner;
-import ToDoListFeature.ToDoItem;
 import ToDoListFeature.ToDoList;
```
adamzzq, line 7: I like how ToDoList related classes are organized in a separate package from the main logic class

src/main/java/ToDoListFeature/ToDoList.java
```diff
@@ -33,83 +52,149 @@ private void unMarkTask(int index) {
...
     public String toString() {
         if (this.toDoListArray.isEmpty()) {
             return "Empty ToDoList";
         }
         String output = "";
         int count = 1;
-        for (ToDoItem item: this.toDoListArray) {
+        for (ToDo item: this.toDoListArray) {
             output += count + "." + item.toString() + "\n";
             count ++;
```
adamzzq: > count ++;

There's no need to leave a space when using an increment operator.
Perhaps use `count++`?

src/main/java/ToDoListFeature/ToDoList.java
```diff
@@ -33,83 +52,149 @@ private void unMarkTask(int index) {
...
+     */
     @Override
     public String toString() {
         if (this.toDoListArray.isEmpty()) {
             return "Empty ToDoList";
         }
         String output = "";
         int count = 1;
-        for (ToDoItem item: this.toDoListArray) {
+        for (ToDo item: this.toDoListArray) {
```
adamzzq: > for (ToDo item: this.toDoListArray) {

Perhaps leave a space before the colon as it is a more conventional coding style, like this:
`for (ToDo item : this.toDoListArray)`

Reference:
[Google Java Style Guide - 4.6.2 Horizontal whitespace: Point number 4 ](https://google.github.io/styleguide/javaguide.html#s4.6.2-horizontal-whitespace)

src/main/java/ToDoListFeature/ToDoList.java
```diff
@@ -33,83 +52,149 @@ private void unMarkTask(int index) {
...
+                    case "quit":
+                        return;
+                
+                    default:
+                        System.out.println("invalid command, use [help] to show all the commands");
+                        break;
+                    }
                 }
+                messageDivider();
+            } catch (StringIndexOutOfBoundsException e) {
```
adamzzq: I like how invalid commands are handled in a cultured way

src/main/java/ToDoListFeature/ToDoList.java
```diff
@@ -33,83 +52,149 @@ private void unMarkTask(int index) {
...
+            int keywordPosition = inputText.indexOf("/by");
+            addItem(new DeadLinesTask(inputText.substring(numberPosition + SPACE_OFFSET, keywordPosition - 1),
+                    inputText.substring(keywordPosition + BY_OFFSET)));
+            return true;
+        } else if (matchesPattern(inputText, EVENTSTASK_PATTERN)) {
+            int keywordPosition1 = inputText.indexOf("/from"); // theck /from keyword position
+            int keywordPosition2 = inputText.indexOf("/to"); // check /to keyword position
+            addItem(new EventsTask(inputText.substring(numberPosition + SPACE_OFFSET, keywordPosition1 - 1),
+                    inputText.substring(keywordPosition1 + FROM_OFFSET, keywordPosition2 - 1),
+                    inputText.substring(keywordPosition2 + TO_OFFSET)));
```
adamzzq: Perhaps can consider dealing with Strings using Matcher methods (since you are already using the regex package), which is more abstract than dealing them with String methods

src/main/java/ToDoListFeature/ToDoList.java
adamzzq, line 1: I like how you abstracted the methods so each method performs one job. Suggestion: perhaps can consider splitting the ToDoList class to better align with the Single responsibility principle

src/main/java/ToDoListFeature/ToDoList.java
```diff
@@ -33,83 +52,149 @@ private void unMarkTask(int index) {
...
         System.out.println(toString());
-        messageDivider();
     }
 
-    public void addItem(ToDoItem item) {
+    /**
+     * Add a ToDo item to the ToDoList
+     * @param item the ToDo item to be added
+     */
+    public void addItem(ToDo item) {
```
adamzzq: An alternative way to name the `ToDo item` would be `task`. It might enhance code readability

src/main/java/ToDoListFeature/DeadLinesTask.java
```diff
@@ -0,0 +1,25 @@
+package ToDoListFeature;
```
adamzzq: Should package name be changed to all lower case?

[reference](https://se-education.org/guides/conventions/java/basic.html#naming)

src/main/java/ChatBBT.java
```diff
@@ -0,0 +1,88 @@
...
+     */
+    public static void featureIntroMessage() {
+        System.out.println("[list] Enter Todo List feature");
+        System.out.println("[bye] Quit");
+        System.out.println("Please enter a command:");
+        messageDivider();
+    }
+
+    public static void createLocalDataFile() {
+        String filePath = "ChatBBTData.txt";
```
Jonoans: Could this be moved out as a class-level constant variable?

src/main/java/ChatBBT.java
```diff
@@ -0,0 +1,88 @@
...
+        try {
+            // Create a File object
+            File file = new File(filePath);
+            // Check if the file already exists
+            if (file.createNewFile()) {
+                System.out.println("File created: " + file.getAbsolutePath());
+                FileWriter fw = new FileWriter("ChatBBTData.txt");
```
Jonoans: Per the comment above, this could reference that constant value so that magic strings can be avoided

src/main/java/todolist/keywordfinder/KeywordPatternMatcher.java
```diff
@@ -0,0 +1,101 @@
...
+    public Keyword getKeywordType() {
+        return keywordType;
+    }
+
+    private int getSpaceCharacterPosition(String inputText) {
+        return inputText.indexOf(" ");
+    }
+
+    public Keyword checkKeywordType() {
+        if (matchesPattern(this.inputText, MARK_FEATURE_PATTERN)) {
```
Jonoans, line 68: Nice use of constant here

src/main/java/todolist/ToDoList.java
```diff
@@ -0,0 +1,239 @@
...
+        StringBuilder dataString = new StringBuilder();
+        for (Task task : this.toDoListArray) {
+            dataString.append(task.storeDataString()).append("\n");
+        }
+        dataString.append("#end");
+        return dataString.toString();
+    }
+
+    public void loadData() {
+        String filePath = "ChatBBTData.txt";
```
Jonoans: This seems to be used across files, maybe you could consider creating a class to contain constants that the program shares across the different source files?

src/main/java/todolist/ToDoList.java
```diff
@@ -0,0 +1,239 @@
...
+        try (BufferedReader reader = new BufferedReader(new FileReader(filePath))) {
+            String line;
+            while (!Objects.equals(line = reader.readLine(), "#end")) {
```
Jonoans: Try to avoid magic strings like that as it may not be apparent what "#end" means

src/main/java/todolist/ToDoList.java
```diff
@@ -0,0 +1,239 @@
...
+                char isDone = line.charAt(2);
+                switch(line.charAt(0)){
+                case 'E' :
```
Jonoans, line 211: Could try to avoid this magic string / character here as well

src/main/java/todolist/ToDoList.java
```diff
@@ -0,0 +1,239 @@
...
+            ErrorMessages.taskDoesNotExistErrorMessage();
+        }
+
+    }
+
+    /**
+     * Execute the ToDoList feature
+     * @param input the scanner to read user input
+     */
+    public void execute(Scanner input) {
```
Jonoans: Try to avoid long methods

src/main/java/todolist/ToDoList.java
Jonoans: Could try to avoid this magic string / character here as well as magic numbers like 0 and 2

Jonoans (commented):
Good job with the program, however try to avoid using magic strings and numbers. Also do try to avoid long methods

PR #61, opened 2024-02-04
[L5-Z] iP

L5-Z:


src/main/java/Jarvas.java
```diff
@@ -0,0 +1,76 @@
...
+        System.out.println("Hello! I'm Jarvas");
+        System.out.println("What can I do for you?");
+        System.out.println(PARTITION_LINE);
+    }
+    public static void terminateBot(){
+        System.out.println(PARTITION_LINE);
+        System.out.println("Bye, see you soon!");
+        System.out.println(PARTITION_LINE);
+    }
+    public static void runTime(){
```
leongxingyu: "Names representing methods must be verbs." 
Can change runTime() to a verb.

src/main/java/Jarvas.java
```diff
@@ -0,0 +1,76 @@
...
+        Task[] tasks = new Task[100];
+        int taskCounter = 0;
+
+        while (true) {
+            if (userInput.equals("bye")) {
+                return;
+            } else if (userInput.startsWith("mark")) {
+                System.out.println(PARTITION_LINE);
+                String index = userInput.substring(5); // obtain value after mark
+                int i = Integer.parseInt(index) - 1; // convert to int
```
leongxingyu: Maybe naming of "i" can be better, but looks ok to me!

src/main/java/Deadline.java
```diff
@@ -0,0 +1,2 @@
+public class Deadline {
```
leongxingyu: Ohno! There isnt any code here! Same for Event.java and ToDo.java

src/main/java/Jarvas.java
```diff
@@ -0,0 +1,76 @@
...
+        System.out.println(" _  | | /'_` )( '__)( ) ( ) /'_` )/',__)");
+        System.out.println("( )_| |( (_| || |   | \\_/ |( (_| |\\__, \\");
+        System.out.println("`\\___/'`\\__,_)(_)   `\\___/'`\\__,_)(____/");
+
+        System.out.println(PARTITION_LINE);
+        System.out.println("Hello! I'm Jarvas");
+        System.out.println("What can I do for you?");
+        System.out.println(PARTITION_LINE);
+    }
+    public static void terminateBot(){
```
leongxingyu: Can add a space before the opening brace. 
e.g 
"public static void terminateBot() {"
Same for runTime() method.

leongxingyu (commented):
Overall, your Jarvas.java looks very neat and readable! There isnt much formatting errors! Can work on your ToDo, Event and Deadline class now! Keep up the good work!

src/main/java/Jarvas.java
```diff
@@ -0,0 +1,104 @@
+import java.util.Scanner;
+public class Jarvas {
+    public static void initialiseBot(){
+        System.out.println(" _____                                  ");
+        System.out.println("(___  )                                 ");
+        System.out.println("    | |   _ _  _ __  _   _    _ _   ___ ");
+        System.out.println(" _  | | /'_` )( '__)( ) ( ) /'_` )/',__)");
+        System.out.println("( )_| |( (_| || |   | \\_/ |( (_| |\\__, \\");
+        System.out.println("`\\___/'`\\__,_)(_)   `\\___/'`\\__,_)(____/");
```
nkotaa: Would it be more readable to define this ASCII art as a string constant with escape characters, then call println() on this variable instead?

src/main/java/Jarvas.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+    public static void terminateBot(){
+        Reply.printReply("Bye, see you soon!");
+    }
+    public static void runTime(){
+        Scanner in = new Scanner(System.in);
+        String userInput = in.nextLine();
+        Task[] tasks = new Task[100];
+
+        while (true) {
```
nkotaa: Is the logic of the loop clear enough to specify an explicit exit condition? Something to the effect of while instruction != "bye".

src/main/java/Jarvas.java
```diff
@@ -0,0 +1,104 @@
...
+                    tasks[i] = deadline;
+                    Reply.printReply(deadline, i+1);
+                    break;
+                }
+                default:
+                    Reply.printReply("Invalid command provided");
+                }
+            } catch (NumberFormatException e) {
+                System.err.println("Invalid command format: " + e.getMessage());
+            }
```
nkotaa: What edge cases does this try / catch block catch that make it necessary? Otherwise it is needlessly introducing an extra indentation to a significant chunk of code.

src/main/java/Reply.java
```diff
@@ -0,0 +1,42 @@
+public class Reply {
+    public static final String PARTITION_LINE = "____________________________________________________________";
+
+    public static void printLine() {
+        System.out.println(PARTITION_LINE);
+    }
+    public static void printReply(String reply) {
+        System.out.println(PARTITION_LINE);
```
nkotaa:
```suggestion
        printLine();
```
And for all further occurrences as well.

src/main/java/Reply.java
```diff
@@ -0,0 +1,42 @@
...
+        System.out.println("You now have " + total + " tasks in the list.");
+        System.out.println(PARTITION_LINE);
+    }
+    public static void printReply(Deadline deadline, int total){
+        System.out.println(PARTITION_LINE);
+        System.out.println("Got it. I've added: ");
+        System.out.println(deadline);
+        System.out.println("You now have " + total + " tasks in the list.");
+        System.out.println(PARTITION_LINE);
+    }
```
nkotaa: Consider refactoring into one method. Since ToDo, Event and Deadline are subclasses of Task, it suffices to define
```
public static void printReply(Task, int)
```
and make use of polymorphism.

src/main/java/Task.java
```diff
@@ -0,0 +1,31 @@
+public class Task {
+    private static int taskCounter = 0;
```
nkotaa:
```suggestion
    private static int numTasks = 0;
```
Do you think this is a clearer name?

nkotaa (commented):
No glaring code readability issues. Aside from excessive nesting in some loops, just a few nits to fix.

src/main/java/Jarvas.java
L5-Z: I think setting it as a constant would require too many line wraps due to length. I have, however moved it to the print class for clarity. Thanks!

src/main/java/Jarvas.java
L5-Z: Yep, line 19 case "bye" is the first case to immediately return the method and move on to terminateBot() to print the exit greeting

src/main/java/Reply.java
L5-Z: Thanks for pointing this out!

src/main/java/Reply.java
L5-Z: Thanks for the correction

src/main/java/Jarvas.java
L5-Z: Thanks for pointing that out!

src/main/java/Jarvas.java
L5-Z: Overlooked this thanks a lot!

src/main/java/Task.java
L5-Z: I think numberOfTasks would be better but a little long so I think I'll keep this thanks!

src/main/java/Jarvas.java
```diff
@@ -0,0 +1,132 @@
...
+        Reply.printArt();
+        Reply.printReply("Hello! I'm Jarvas", "What can I do for you?");
+    }
+    public static void terminateBot() {
+        Reply.printReply("Bye, see you soon!");
+    }
+    public static void processInput() {
+        Scanner in = new Scanner(System.in);
+        String userInput = in.nextLine();
+        Task[] tasks = new Task[100];
```
json-bot-boop: Try to avoid magic numbers

src/main/java/Jarvas.java
```diff
@@ -0,0 +1,132 @@
...
+            String instruction = userInput.split(" ")[0];
+            try {
+                switch (instruction) {
+                case "bye": {
+                    return;
+                }
+                case "mark": {
+                    String index = userInput.substring(4).trim(); // obtain value after mark
+                    if (index.isEmpty()) {
+                        throw new CustomException("Index to mark done is unspecified.");
```
json-bot-boop: Try to avoid magic strings

src/main/java/Jarvas.java
```diff
@@ -0,0 +1,132 @@
...
+                    if (deadlineDetails.isEmpty()) {
+                        throw new CustomException("Deadline cannot be left blank.");
+                    }
+                    String[] deadlineParameters = Deadline.getDeadline(deadlineDetails);
+                    String deadlineLabel = deadlineParameters[0];
+                    Deadline deadline = new Deadline(deadlineLabel, deadlineParameters[1]);
+
+                    tasks[taskCount] = deadline;
+                    Reply.printReply(deadline, taskCount+1);
+                    break;
```
json-bot-boop: Can consider the case where the argument values are empty

src/main/java/Reply.java
```diff
@@ -0,0 +1,84 @@
...
+        System.out.println("deadline - adds a deadline to the List");
+        System.out.println("mark - indicates an item on the List as done");
+        System.out.println("unmark - indicates an item on the List as not done");
+    }
+    public static void printManual(String command) {
+        printLine();
+        try {
+            switch (command) {
+            case "": {
+                System.out.println("man format: man *command*");
```
json-bot-boop: Maybe can consider putting the string in println as a constant variable instead

src/main/java/Jarvas.java
```diff
@@ -0,0 +1,132 @@
...
+                }
+                case "list":
+                    if (Task.enumerateTask() == 0) {
+                        throw new CustomException("List is empty.");
+                    }
+                    System.out.println("Here are the tasks in your list:");
+                    for (int i = 0; i < Task.enumerateTask(); i++) {
+                        System.out.println((i + 1) + ". "  + tasks[i]);
+                    }
+                    break;
```
json-bot-boop: Can consider abstracting methods to not make the current method very long

src/main/java/Event.java
```diff
@@ -0,0 +1,36 @@
...
+        super(label);
+        this.from = from;
+        this.to = to;
+    }
+
+    public static String[] getInterval(String input) {
+
+        String[] results = new String[3];
+
+        // Find the index of "/from" and "/to"
```
json-bot-boop: Try not repeat in comments information that is obvious from the code. This code quality will be covered in future lectures.

src/main/java/Reply.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+
+    public static void printReply(String reply) {
+        printLine();
+        System.out.println(reply);
+        printLine();
+    }
+
+    public static void printReply(String s1, String s2) {
```
json-bot-boop: Maybe can use another variable name other than `s1` and `s2`

json-bot-boop (commented):
Overall, it is ok except there is some code quality and coding standard issue.

src/main/java/Jarvas.java
L5-Z: Empty parameters are considered in deadlineDetails.isEmpty 🫡

L5-Z: Or are you referring to the /to string

json-bot-boop: Yeap, I am referring to the `/to` value

PR #69, opened 2024-02-04
[Zheqi] ip

adamzzq:


src/main/java/Adam.java
```diff
@@ -0,0 +1,25 @@
...
+            try {
+                CommandGenerator.generate(input.nextLine())
+                        .orElseThrow()
+                        .execute(tasks);
+            } catch (NoSuchElementException error) {
+                System.out.println(Message.ERROR_MESSAGE);
+            }
+
+            System.out.println(Message.DELIMITER);
+        }
```
dylansiew, line 44: Good use of constants, remember to close the scanner

src/main/java/Command.java
```diff
@@ -0,0 +1,3 @@
+public interface Command {
```
dylansiew: Good use of abstractions

src/main/java/TaskGenerator.java
```diff
@@ -0,0 +1,10 @@
+public class TaskGenerator {
+    public static Task generate(String[] parsedInput) {
+        if (parsedInput.length == 1) {
+            return new ToDo(parsedInput[0]);
+        } else if (parsedInput.length == 2) {
```
dylansiew: Hardcoding to the length of the parsedInput could possibly cause issues in the future when adding more Tasks

src/main/java/Message.java
```diff
@@ -0,0 +1,57 @@
...
+            + GREETING;
+
+    public static final String EXIT_MESSAGE = "Farewell, adventurer! Until our paths cross again!\n" + DELIMITER;
+
+    public static final String MARK_MESSAGE = "Ta-da! Task conquered! I've marked this task as done:\n";
+
+    public static final String UNMARK_MESSAGE = "Whoopsie-daisy! This task is back in action:\n";
+
+    public static final String ERROR_MESSAGE = "Oopsie! Looks like I got tangled up in my circuits.\n"
+            + "Could you please try again?";
```
dylansiew: Good implementation of constants

dylansiew (commented):
Overall very understandable code

src/main/java/Parser.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public enum Parser {
+    BYE("bye"),
+    LIST("list"),
+    TOGGLE("^(mark [0-9]*|unmark [0-9]*)"),
+    TODO("todo\\s(.+)"),
+    DEADLINE("deadline\\s(.+)/by\\s(.+)"),
+    EVENT("event\\s(.+)/from\\s(.+)/to\\s(.+)");
```
joshualeejunyi: I like the use of enum to make it more readable!

src/main/java/Adam.java
```diff
@@ -0,0 +1,25 @@
+import java.util.NoSuchElementException;
+import java.util.Scanner;
+
+public class Adam {
+    public static void main(String[] args) {
+        Scanner input = new Scanner(System.in);
```
joshualeejunyi: Don't forget to close the scanner at the end so that you don't leak resources!

src/main/java/CommandGenerator.java
```diff
@@ -0,0 +1,19 @@
+import java.util.Optional;
+
+public class CommandGenerator{
+    public static Optional<Command> generate(String input) {
+        switch (Parser.analyzeInput(input)) {
```
joshualeejunyi: Just a nitpick, but might be better to split lines into more readable chunks. Noticed the same across the files as well :)
Something like: 
```suggestion
        Parser parsedInput = Parser.analyzeInput(input);
        switch (parsedInput) {
```

src/main/java/Deadline.java
```diff
@@ -0,0 +1,13 @@
...
+    protected String by;
+
+    public Deadline(String description, String by) {
+        super(description);
+        this.by = by;
+    }
+
+    @Override
+    public String toString() {
+        return "[D]" + super.toString() + " (by: " + by + ")";
```
joshualeejunyi: Minor nitpick, but why not String.format() if you have used it in Event.java? Would be great to standardize your code.

src/main/java/Message.java
```diff
@@ -0,0 +1,57 @@
+import java.util.Random;
+
+public class Message {
```
joshualeejunyi: Really like this class of constants! I was considering something similar for my own iP as well

src/main/java/TaskGenerator.java
```diff
@@ -0,0 +1,10 @@
...
+            return new Deadline(parsedInput[0], parsedInput[1]);
+        }
+        return new Event(parsedInput[0], parsedInput[1], parsedInput[2]);
+    }
```
joshualeejunyi: Perhaps you can reuse enums that you have defined above to determine the type of task that you are generating?

src/main/java/TaskList.java
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    public int size() {
+        return this.tasks.size();
+    }
+
+    public void displayAll() {
+        IntStream.rangeClosed(1, tasks.size())
+                .mapToObj(index -> index + "." + getTask(index))
+                .forEach(System.out::println);
```
joshualeejunyi: I would suggest splitting these into multiple lines, or commenting on each line so that it will be easy to remember what it is actually doing!

joshualeejunyi (commented):
Overall, good effort! Some tweaking and cleaning up might be useful to make things easier to read and understand :)

src/main/java/Message.java
```diff
@@ -0,0 +1,58 @@
...
+    public static final String DELIMITER = "\n>====> >====> >====> >====> >====> >====> >====> >====> >====>\n";
+
```
yuhengr: This is a great way to keep constants organised!

src/main/java/Parser.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.regex.Pattern;
+
+public enum Parser {
+    BYE("bye"),
+    LIST("list"),
+    TOGGLE("(mark|unmark)\\s(\\d+)"),
+    TODO("todo\\s(.+)"),
+    DEADLINE("deadline\\s(.+)/by\\s(.+)"),
+    EVENT("event\\s(.+)/from\\s(.+)/to\\s(.+)");
+
```
yuhengr: Great work handling the inputs in an organised way

src/main/java/Message.java
```diff
@@ -0,0 +1,58 @@
...
+            + "╚═╝  ╚═╝╚═════╝ ╚═╝  ╚═╝╚═╝     ╚═╝\n";
+
+    private static final String LOGO_3D = " ________  ________  ________  _____ ______      \n"
+            + "|\\   __  \\|\\   ___ \\|\\   __  \\|\\   _ \\  _   \\    \n"
+            + "\\ \\  \\|\\  \\ \\  \\_|\\ \\ \\  \\|\\  \\ \\  \\\\\\__\\ \\  \\   \n"
+            + " \\ \\   __  \\ \\  \\ \\\\ \\ \\   __  \\ \\  \\\\|__| \\  \\  \n"
+            + "  \\ \\  \\ \\  \\ \\  \\_\\\\ \\ \\  \\ \\  \\ \\  \\    \\ \\  \\ \n"
+            + "   \\ \\__\\ \\__\\ \\_______\\ \\__\\ \\__\\ \\__\\    \\ \\__\\\n"
+            + "    \\|__|\\|__|\\|_______|\\|__|\\|__|\\|__|     \\|__|\n";
+
```
yuhengr: Perhaps the lineSeparator method could be used here? I believe it was mentioned that the behaviour of \n depends on the platform. Please correct me if I'm mistaken

src/main/java/Parser.java
```diff
@@ -0,0 +1,63 @@
...
+        Pattern eventPattern = Pattern.compile(EVENT.getCommandRegex());
+        Pattern deadlinePattern = Pattern.compile(DEADLINE.getCommandRegex());
+        Pattern todoPattern = Pattern.compile(TODO.getCommandRegex());
+        Pattern togglePattern = Pattern.compile(TOGGLE.getCommandRegex());
+
+        Matcher eventMatcher = eventPattern.matcher(input);
+        Matcher deadlineMatcher = deadlinePattern.matcher(input);
+        Matcher todoMatcher = todoPattern.matcher(input);
+        Matcher toggleMatcher = togglePattern.matcher(input);
+
```
yuhengr: Perhaps having some comments here for explanation will help as it seems like there is quite a bit of things happening

yuhengr (commented):
Overall great work in making use of OOP to abstract the different classes and ensuring that the code is well organised. Keep up the good work

src/main/java/ui/Message.java
```diff
@@ -0,0 +1,98 @@
+package ui;
+
+import java.util.Random;
+
+public class Message {
```
Jonoans, line 8: Good job with the use of a constants class :D

Jonoans (commented):
Really well done! Not much issues. Good use of constants. Methods are short and easy to digest as individual units which helps one to better understand how code units work together to compose functionalities, good job!

PR #70, opened 2024-02-04
[CerealMiller] iP

CerealMiller:
Completed up to iP Level-3 activities.


src/main/java/Duke.java
```diff
@@ -1,10 +1,62 @@
...
+        String logo = "  _________   _____      _____   \n" +
+                " /   _____/  /  _  \\    /     \\  \n" +
+                " \\_____  \\  /  /_\\  \\  /  \\ /  \\ \n" +
+                " /        \\/    |    \\/    Y    \\\n" +
+                "/_______  /\\____|__  /\\____|__  /\n" +
+                "        \\/         \\/         \\/ ";
+        System.out.println("Hello! I'm SAM\n" + logo + "\n" + "What can I do for you?\n");
+
+        // Initializing scanner to read user input
+        Scanner in = new Scanner(System.in);
```
ttasnim5: This is great! I think you can maybe benefit from using BufferedReader instead of Scanner because Scanner tends to be very slow in practice.

src/main/java/Duke.java
```diff
@@ -1,10 +1,62 @@
...
+        String line = in.nextLine();
+
+        // Array to store tasks
+        Task[] records = new Task[100];
```
ttasnim5: Great job pluralizing the array

src/main/java/Duke.java
```diff
@@ -1,10 +1,62 @@
...
+        int numItems = 0; // Number of tasks
+        int listIndex; // Index for tasks
+
+        // Main loop
+        while (!line.equals("bye")) {
+            String[] words = line.split(" ");
```
ttasnim5: You may need to incorporate exception handling because you are taking user input

src/main/java/Task.java
```diff
@@ -0,0 +1,22 @@
+public class Task {
+    // Fields to store task information
+    protected String description; // Description of the task
+    protected boolean isDone; // Flag to indicate if the task is done
```
ttasnim5: Good naming sense for the boolean

src/main/java/Sam.java
```diff
@@ -0,0 +1,129 @@
...
+                line = in.nextLine();
+            } catch (InvalidInputException e) {
+                System.out.println("Try again!");
+                printLine();
+                line = in.nextLine(); // Prompt for new input
+            } catch (NumberFormatException e) {
+                System.out.println("That's not a number!");
+                printLine();
+                line = in.nextLine(); // Prompt for new input
+            } catch(IndexOutOfBoundsException e){
```
madhanse: There needs to be a space between 'catch' and (, and, ) and {, as per coding standards

src/main/java/Sam.java
```diff
@@ -0,0 +1,129 @@
...
+                    System.out.println("Nice! I've marked this task as done:\n" + records[listIndex]);
+                    break;
+                case "unmark":
+                    // Marking a task as not done
+                    listIndex = Integer.parseInt(words[1]) - 1;
+                    markCheck(listIndex, numItems);
+                    records[listIndex].isDone = false;
+                    System.out.println("OK, I've marked this task as not done yet:\n" + records[listIndex]);
+                    throw new InvalidInputException();
+                    //break;
```
madhanse: Need to add '\\Fallthrough' comment if there is no break statement inside the case block, as per coding standards
image

PR #77, opened 2024-02-05
[Eric Sim Chong Li] iP

e0958902:


src/main/java/Lotes.java
```diff
@@ -0,0 +1,108 @@
...
+                + "     added: " + description + lineSeparator
+                + horizontalLine + lineSeparator);
+    }
+
+    public static void main(String[] args) {
+        String logo = "  #        ####  ##### ######  ####\n"
+                + "                #       #    #   #   #      #\n"
+                + "                #       #    #   #   #####   ####\n"
+                + "                #       #    #   #   #           #\n"
+                + "                #######  ####    #   ######  ####";
```
e0958902, line 79: I like how you separated the lines to ensure readability,

e0958902 (commented):
This is just a test comment.

src/main/java/Task.java
```diff
@@ -0,0 +1,21 @@
+public class Task {
+    protected String description;
```
annoy-o-mus: What kind of description? Can be more descriptive like `taskDescription`

src/main/java/Lotes.java
```diff
@@ -0,0 +1,108 @@
...
+        Scanner inputCommand = new Scanner(System.in); // Prompt for continuous user input.
+
+        while (true) {
+            String userInput = inputCommand.nextLine();
+
+            if (userInput.equals("bye")) {
+                System.out.println(horizontalLine + lineSeparator
+                        + "    Bye. Hope to see you again soon!\n" + horizontalLine);
+                break; // Exit the program
+            } else if (userInput.equals("list")) {
```
annoy-o-mus: Can consider a switch statement instead of multiple if-else statements

src/main/java/Lotes.java
```diff
@@ -0,0 +1,108 @@
+import java.util.Scanner;
+import java.util.ArrayList;
+public class Lotes {
+    private static final ArrayList<Task> taskList = new ArrayList<>();
+    private static final String lineSeparator = System.lineSeparator();
+    private static final String horizontalLine = "    _________________________"
+                                                + "___________________________";
+
+    private static void printTasksList() {
```
annoy-o-mus: inclusion of javadocs will be great

src/main/java/Lotes.java
```diff
@@ -0,0 +1,108 @@
...
+        System.out.println(horizontalLine);
+
+        if (taskList.isEmpty()) {
+            System.out.println("     List is empty, please enter some text to add to list.");
+        } else {
+            System.out.println("     Here are the tasks in your list:");
+
+            int index = 0;
+            for (Task task : taskList) {
+                index++;
```
annoy-o-mus: What is the index referring

src/main/java/Lotes.java
```diff
@@ -0,0 +1,108 @@
...
+
+            taskList.get(taskListIndex).setDone(false);
+
+            System.out.print(horizontalLine + lineSeparator
+                    + "     OK, I've marked this task as not done yet:" + lineSeparator
+                    + "     [" + taskList.get(taskListIndex).getStatusIcon() + "] "
+                    + taskList.get(taskListIndex).getDescription() + lineSeparator
+                    + horizontalLine + lineSeparator);
+
+        } catch (IndexOutOfBoundsException | NumberFormatException s) {
```
annoy-o-mus: Suggestion: Can consider separating the exceptions into separate catch staments so that you can better handle each exception individually. 

Alternatively, you can consider scoping the tasks into its own try-catch block so that the scope of exception is limited to that task. For example, instead of 

``` java
try {
    // Code snippet "a" 
    // Code snippet "b"
} catch (/* Exceptions thrown by "a" and/or "b" */) {
    // Catch Code
}
```


you can do

``` java
try {
    // Code snippet "a" 
} catch (/* Exceptions thrown by "a" */) {
    // Catch Code
}
// Some code
try {
    // Code snippet "b"
} catch (/* Exceptions thrown by "b" */){ 
    // Catch Code
}
```

src/main/java/Task.java
```diff
@@ -0,0 +1,21 @@
...
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public String getDescription() {
+        return description;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); // returns X if task isDone is true
```
MonkeScripts: Perhaps try placing mark and unmark functionalities within the task class itself. One example would be to place the umark/mark methods within the task class itself when marking/unmarking a task. 
This makes it easier to debug because everything is contained within the task class itself rather than having an external setDone() method

src/main/java/Lotes.java
```diff
@@ -0,0 +1,108 @@
+import java.util.Scanner;
+import java.util.ArrayList;
+public class Lotes {
+    private static final ArrayList<Task> taskList = new ArrayList<>();
+    private static final String lineSeparator = System.lineSeparator();
+    private static final String horizontalLine = "    _________________________"
```
MonkeScripts: Perhaps it is better to create a separate method to print newlines
```
private static void printLine(int numLines) {
    System.out.println("_" * numLines);
 }
```

src/main/java/Lotes.java
```diff
@@ -0,0 +1,108 @@
...
+
+        if (taskList.isEmpty()) {
+            System.out.println("     List is empty, please enter some text to add to list.");
+        } else {
+            System.out.println("     Here are the tasks in your list:");
+
+            int index = 0;
+            for (Task task : taskList) {
+                index++;
+                System.out.print("     " + index + ".[" + task.getStatusIcon()
```
MonkeScripts: Perhaps try Overriding the toString() method in task to reduce the number of lines here

src/main/java/Lotes.java
```diff
@@ -0,0 +1,108 @@
...
+            printTasksList();
+        }
+    }
+
+    private static void unMarkTask(String userInput) {
+        try {
+            int taskListIndex = (Integer.parseInt(userInput.substring(7))
+                    - 1);
+
+            taskList.get(taskListIndex).setDone(false);
```
MonkeScripts: Maybe add some comments!

src/main/java/lotes/Lotes.java
```diff
@@ -0,0 +1,22 @@
+package lotes;
+
+import lotes.parser.Parser;
+import lotes.task.TaskList;
+import lotes.storage.Storage;
+import java.util.Scanner;
+
+public class Lotes {
+
+    public static final Scanner inputCommand = new Scanner(System.in); // Setting the Scanner
```
okkhoy: redundant comment

src/main/java/lotes/parser/Parser.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.Scanner;
+
+import lotes.storage.Storage;
+import lotes.task.TaskList;
+
+public class Parser {
+
+    // Parses the user input to get the command to perform.
+
+    public static String[] parseInputCommand(String userInput) {
```
okkhoy: format of header comment is incorrect. pl write it as javadoc comments

src/main/java/lotes/parser/Parser.java
```diff
@@ -0,0 +1,107 @@
...
+            break;
+        default:
+            if(command.length < 2) {
+                throw new LotesException();
+            }
+        }
+
+        if (command[0].equals("todo") || command[0].equals("deadline") || command[0].equals("event")
+        || command[0].equals("add") || command[0].equals("mark") || command[0].equals("unmark")
+        || command[0].equals("delete")) {
```
okkhoy: isn't it easier to check for which command don't need updating and skipping it?

src/main/java/lotes/parser/Parser.java
```diff
@@ -0,0 +1,107 @@
...
+
+    // Handle errors by catching exceptions.
+
+    public static boolean isException(String userInput, TaskList taskList) {
+        try {
+            if (Parser.performUserInput(userInput, taskList)) { // if ixExit returns true
+                return true;
+            }
+
+        } catch (NumberFormatException e) {
```
okkhoy: Naming and comments are a bit strange.
You are actually executing the command here and catching the exception if one occurs.
However, the name gives an impression of a simple boolean check

src/main/java/lotes/parser/Parser.java
```diff
@@ -0,0 +1,107 @@
+package lotes.parser;
+
+import java.util.Scanner;
+
+import lotes.storage.Storage;
+import lotes.task.TaskList;
+
+public class Parser {
```
okkhoy, line 18: Parser is doing both parsing and executing the command. in iP this is kinda OK. However, you are violating some design principles you will learn about soon. Avoid this in tP

src/main/java/lotes/storage/Storage.java
```diff
@@ -0,0 +1,76 @@
...
+    // Update the current file with the updated task list
+
+
+    public static void updateFile(){
+        try {
+            FileWriter fw = new FileWriter(filename);
+            StringBuilder sb = new StringBuilder();
+//            for(int i = 0; i < TaskList.getTaskCount(); i++) {
+//                sb.append(TaskList.taskList[i]).append(TaskList.separator);
+//            }
```
okkhoy: pls remove dead code

src/main/java/lotes/task/Task.java
```diff
@@ -0,0 +1,36 @@
+package lotes.task;
+
+public class Task {
```
okkhoy, line 3: need some documentation for the Task or its sub classes.

okkhoy (commented):
I hope these review comments help in polishing the iP code before submission.

PR #79, opened 2024-02-05
[Tay Jun Feng Vernon] iP 

V4Vern:


src/main/java/Byte.java
```diff
@@ -0,0 +1,116 @@
+import java.util.Scanner;
+
+public class Byte {
+    private static final int MAX_TASKS = 100;
+    private static final Task[] tasks = new Task[MAX_TASKS];
+    private static int taskCount = 0;
+
```
ymirmeddeb: unnecessary white space

src/main/java/Byte.java
```diff
@@ -0,0 +1,116 @@
...
+                break;
+            }
+        }
+    }
+
+    public static boolean processUserInput(String userInput){
+        if (userInput.equals("bye")) {
+            printGoodbyeMessage();
+            return true;
+        } else if (userInput.startsWith("mark ")) {
```
ymirmeddeb: else if should be on a new line after brackets

src/main/java/Byte.java
```diff
@@ -0,0 +1,116 @@
...
+            taskNumber = Integer.parseInt(taskNumberStr)-1;
+            Task task = tasks[taskNumber];
+            if (isDone) {
+                task.markAsDone();
+                System.out.println("Nice! I've marked this task as done:\n  " + tasks[taskNumber]);
+            } else {
+                task.markAsNotDone();
+                System.out.println("OK, I've marked this task as not done yet:\n  " + tasks[taskNumber]);
+            }
+        } catch (NumberFormatException e) {
```
ymirmeddeb: catch should be on a new line after brackets

src/main/java/Byte.java
```diff
@@ -0,0 +1,116 @@
...
+            System.out.println("Please enter a valid task number.");
+        } catch (IndexOutOfBoundsException e) {
```
ymirmeddeb: parameter e should be named something different, not very clear what it does

src/main/java/ToDo.java
```diff
@@ -0,0 +1,12 @@
...
+
+    public ToDo(String description) {
+        super(description);
+    }
+
+    @Override
+    public String toString(){
+        return "[T]" + super.toString();
+    }
+}
```
ymirmeddeb: overall, code is very well written and follows coding standard for the most part. great job!

src/main/java/Byte.java
```diff
@@ -0,0 +1,116 @@
+import java.util.Scanner;
+
+public class Byte {
+    private static final int MAX_TASKS = 100;
```
rexyyong: I like that you used a constant to indicate MAX_TASKS, which makes it clear to readers that the number 100 represents max tasks

src/main/java/Byte.java
```diff
@@ -0,0 +1,116 @@
...
+            return true;
+        } else if (userInput.startsWith("mark ")) {
+            markTask(userInput.substring("mark ".length()), true);
+        } else if (userInput.startsWith("unmark ")) {
+            markTask(userInput.substring("unmark ".length()), false);
+        } else if (userInput.equals("list")) {
+            listTasks();
+        } else if (userInput.startsWith("todo ")){
+            handleToDoCommand(userInput);
+        }else if(userInput.startsWith("deadline ")){
```
rexyyong: Remember to include the space between } and else if, as well as the space between if and (....)

src/main/java/Byte.java
```diff
@@ -0,0 +1,116 @@
...
+        printLineSeparator();
+    }
+
+    private static void addTask(Task task) {
+        if (taskCount < MAX_TASKS) {
+            tasks[taskCount++] = task;
+            System.out.println("Got it. I've added this task:\n " + task);
+            System.out.println("Now you have " + taskCount + " tasks in the list.");
+        } else {
+            System.out.println("Sorry, I can only handle up to " + MAX_TASKS + " tasks!");
```
rexyyong: I like that you included a case where more than 100 tasks are added, which prevents users from adding more than 100 tasks

text-ui-test/input.txt
```diff
@@ -0,0 +1,3 @@
+todo borrow book
+list
+bye
```
rexyyong, line 11: Test case is well done, consider including mark, unmark and list to test if mark and unmark is done correctly

src/main/java/Byte.java
```diff
@@ -0,0 +1,116 @@
...
+            handleToDoCommand(userInput);
+        }else if(userInput.startsWith("deadline ")){
+            handleDeadlineCommand(userInput);
+        }else if(userInput.startsWith("event ")){
+            handleEventCommand(userInput);
+        }
+        return false;
+    }
+
+    private static void handleToDoCommand(String userInput) {
```
rexyyong: Is there a reason why you didn't put handleToDoCommand function under class ToDo, I noticed this for handleDeadline etc

src/main/java/Byte.java
```diff
@@ -0,0 +1,116 @@
...
+    private static void listTasks() {
+        for (int i = 0; i < taskCount; i++) {
+            System.out.println((i + 1) + ". " + tasks[i]);
+        }
+        printLineSeparator();
+    }
+
+    private static void markTask(String taskNumberStr, boolean isDone) {
+        int taskNumber;
+        try {
```
rexyyong: I like that you implemented try and catch errors here to handle cases where task number is not keyed in correctly

src/main/java/Byte.java
```diff
@@ -0,0 +1,116 @@
...
+        } catch (NumberFormatException e) {
+            System.out.println("Please enter a valid task number.");
+        } catch (IndexOutOfBoundsException e) {
+            System.out.println(e.getMessage());
+        }
+        printLineSeparator();
+    }
+
+    private static void printLineSeparator() {
+        System.out.println("____________________________________________________________");
```
rexyyong: consider creating a function named printLine, which allows you to print the long line everytime you need to print such a long line

rexyyong (commented):
Overall I liked the function namings. There were some very minor formatting errors which aren't too jarring, most of which are already mentioned by the other reviewer. 

However I think it may be a good idea to shift some of the functions to the other classes (Deadline, Event, Task etc) to make your Byte.java more readable.

src/main/java/Byte.java
V4Vern: I followed according to "https://se-education.org/guides/conventions/java/basic.html" though.

src/main/java/Byte.java
V4Vern: Same! I followed according to "https://se-education.org/guides/conventions/java/basic.html" though.

src/main/java/Byte.java
V4Vern: I thought it's a widely recognized practice in Java to use e or ex as the exception variable and the exception mentioned the error so I assumed I don't have to rename it.

src/main/java/Byte.java
V4Vern: wait I'm confused? do you mean to rename the method name cause this function prints long lines matches what you described.

src/main/java/Byte.java
V4Vern: Task classes remain dedicated to modelling tasks, not parsing user input or handling application commands which is one of the reasons why I didn't put it under the class but rather on the main method.

src/main/java/Byte/Byte.java
```diff
@@ -0,0 +1,163 @@
...
+            throw new ByteException("I'm sorry, but I don't know what that means.");
+        }
+        return false;
+    }
+
+    private static void handleToDoCommand(String userInput) throws ByteException {
+        String description = userInput.substring("todo ".length()).trim();
+        if (description.isEmpty()){
+            throw new ByteException("The description of a todo cannot be empty.");
+        }
```
yicheng-toh: Good use of guard clause to ensure happy path.

src/main/java/Byte/Byte.java
```diff
@@ -0,0 +1,163 @@
...
+
+    private static void handleEventCommand(String userInput) throws ByteException {
+        String trimmedInput = userInput.trim();
+        if (!trimmedInput.contains(" /from ") || !trimmedInput.contains(" /to ")) {
+            throw new ByteException("Oops! The event command is missing a '/from' and '/to' to specify the event timing.");
+        }
+        String descriptionPart = trimmedInput.split(" /from ", 2)[0];
+        String timingPart = trimmedInput.split(" /from ", 2)[1];
+        String description = descriptionPart.substring("event ".length()).trim();
+        String[] timingDetails = timingPart.split(" /to ", 2);
```
yicheng-toh: I like how you named the variables.

src/main/java/Byte/Byte.java
```diff
@@ -0,0 +1,266 @@
...
+    private static final int EVENT_END_TIME_INDEX = 4;
+
+
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+        runByte(scanner);
+        if (tasksChanged){
+            try{
+                saveTasksToFile();
+            }catch (IOException e){
```
yicheng-toh:
```suggestion
            } catch (IOException e) {
```
Perhaps considering adding spaces in accordance to [java coding standard](https://se-education.org/guides/conventions/java/basic.html).

src/main/java/Byte/Byte.java
```diff
@@ -0,0 +1,266 @@
...
+    private static void processTaskFromLine(String line) {
+        String[] parts = line.split(" \\| ");
+        String type = parts[TASK_TYPE_INDEX];
+        String description = parts[DESCRIPTION_INDEX];
+        switch (type) {
+            case "T":
+                addTask(new ToDo(description));
+                break;
+            case "D":
+                String by = parts[DEADLINE_TIME_INDEX ];
```
yicheng-toh:
```suggestion
                String by = parts[DEADLINE_TIME_INDEX];
```
Perhaps delete the space here to be consistent with other part of the code.

src/main/java/Byte/Byte.java
```diff
@@ -0,0 +1,266 @@
...
+    private static final int DESCRIPTION_INDEX = 2;
+    private static final int DEADLINE_TIME_INDEX = 3;
+    private static final int EVENT_START_TIME_INDEX = 3;
+    private static final int EVENT_END_TIME_INDEX = 4;
+
+
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+        runByte(scanner);
+        if (tasksChanged){
```
yicheng-toh: You may want to consider having code of similar abstraction together. [SLAP](https://se-education.org/guides/conventions/java/basic.html).

yicheng-toh (commented):
LGTM. Please keep up the standard.

src/main/java/Byte/Byte.java
V4Vern: Noted

src/main/java/Byte/Byte.java
V4Vern: Noted

src/main/java/Byte/Byte.java
V4Vern: Noted

PR #84, opened 2024-02-05
[nikhil-2101] iP

nikhil-2101:


src/main/java/Duke.java
```diff
@@ -19,20 +21,90 @@ public static void main(String[] args) {
...
+                }
+
+                String description = parts[0].trim();
+                String from = parts[1].trim();
+                String to = parts[2].trim();
+
+                Event newEvent = new Event(description, from, to);
+                tasks[taskCount] = newEvent;
+                taskCount++;
+
```
HikoHikoHiko: In my opinion, here is a bit long nesting. Some conditions can be separated out as an isolated else if statement.

src/main/java/Duke.java
```diff
@@ -19,20 +21,90 @@ public static void main(String[] args) {
...
+                Todo newTodo = new Todo(description);
+                tasks[taskCount] = newTodo;
+                taskCount++;
+
+                System.out.println("____________________________________________________________");
+                System.out.println("Got it. I've added this task:");
+                System.out.println("  " + newTodo);
+                System.out.println("Now you have " + taskCount + " tasks in the list.");
+                System.out.println("____________________________________________________________");
+            } else if (userInput.startsWith("deadline ")) {
```
HikoHikoHiko: Also, adopt 'KISSing', make your code simple. For example, you can isolate out some arguments as isolated class file to avoid long winded function in the main class file.

src/main/java/Duke.java
```diff
@@ -19,20 +21,90 @@ public static void main(String[] args) {
...
+                int index = content.indexOf("/by");
+
+                if (index != -1) {
+                    System.out.println("Please use '/by' to specify the deadline time.");
+                    String description = content.substring(0, index).trim();
+                    String by = content.substring(index + 4).trim();
+                    Deadline newDeadline = new Deadline(description, by);
+                    tasks[taskCount] = newDeadline;
+                    taskCount++;
+                    System.out.println("____________________________________________________________");
```
HikoHikoHiko: Here, instead of using 'system.out.println" with a long divider, you can define it at the start of the file, To make it simple by just simplying calling println(breakoutline) etc.

src/main/java/Duke.java
```diff
@@ -8,7 +8,9 @@ public static void main(String[] args) {
         System.out.println("____________________________________________________________");
 
         Scanner scanner = new Scanner(System.in);
-        ArrayList<Task> tasks = new ArrayList<>();
+        Task[] tasks = new Task[100];
```
HikoHikoHiko: Instead of limiting the max size of task to be 100. Maybe what you can do is to hash the array to a bigger array whenever the task no. goes beyond 100. i.e. add the rehash function.

HikoHikoHiko (commented):
Generally aligns with the code quality standard. Just in my opinion I feel that having a long if/else statement may lead to slowing down of the programme? May be using switch would be better? 
Good effort and well done.

PR #93, opened 2024-02-05
[nigelheng] iP

nigelheng:
Replace Duke with Eln in source file
Create echoUser method to repeat user input
Update echoUser method to add user input into list that can be printed


src/main/java/Eln.java
```diff
@@ -0,0 +1,47 @@
+
+import java.util.Scanner;
+
+
+public class Eln {
+    static String LINE = "____________________________________________________________";
```
sxfoo: Maybe consider making this private (if it wont be used out of this class) and final (constant)?
```suggestion
    private static final String LINE = 
    "____________________________________________________________";
```

src/main/java/Eln.java
```diff
@@ -0,0 +1,47 @@
...
+    private static void echoUser() {
```
sxfoo: The method name "echoUser" does not match this method's function. Perhaps rename it to something else?
```suggestion
    private static void handleCommands() {
```

src/main/java/Eln.java
```diff
@@ -0,0 +1,47 @@
...
+                System.out.println(LINE);
+            } else {
+                list[index] = input;
+                System.out.println(LINE);
+                System.out.println("added: " + input);
+                System.out.println(LINE);
+                index++;
+            }
+        }
+    }
```
sxfoo: Good note in using K&R style brackets! (aka Egyptian style) for if else statements

src/main/java/Eln.java
sxfoo: Besides that, I think that you made good note to use capital letters in "LINE" to use for this constant. Nice!

src/main/java/Eln.java
```diff
@@ -0,0 +1,47 @@
...
+
+public class Eln {
+    static String LINE = "____________________________________________________________";
+    private static void echoUser() {
+        Scanner scan = new Scanner(System.in);
+        String input = "";
+        String[] list = new String[20];
+        int index = 0;
+
+        while(!input.equals("bye")) {
```
sxfoo: Do remember to include a space between while and the condition!
I noticed this happening for your if and for loop statements down below also
```suggestion
        while (!input.equals("bye")) {
```

src/main/java/Eln.java
```diff
@@ -0,0 +1,47 @@
+
+import java.util.Scanner;
+
+
+public class Eln {
+    static String LINE = "____________________________________________________________";
+    private static void echoUser() {
+        Scanner scan = new Scanner(System.in);
+        String input = "";
+        String[] list = new String[20];
```
sxfoo: I like that you kept the array specifiers to the String type and not list!

sxfoo (commented):
Generally looks good coding standard wise!
Some cosmetic changes requested.

src/main/java/Eln.java
jthibault03, line 1: good how you put LINE in all caps to avoid the "magic string" in your code

src/main/java/Eln.java
```diff
@@ -0,0 +1,47 @@
...
+                if(index == 0) {
+                    System.out.println("list is empty");
+                }
+                for(int i = 0; i < index; i++) {
+                    System.out.println((i+1) + ". " + list[i]);
+                }
+                System.out.println(LINE);
+            } else {
+                list[index] = input;
+                System.out.println(LINE);
```
jthibault03: I think it would be good if you created a private method to handle printing different segments, to avoid typing System.out.println(LINE) many times in your code and to create a level of abstraction.

src/main/java/Eln.java
```diff
@@ -0,0 +1,47 @@
...
+        }
+    }
+    public static void main(String[] args) {
+
+        System.out.println(LINE);
+        System.out.println("Hello! I'm Eln");
+        System.out.println("What can I do for you?");
+        System.out.println(LINE);
+
+        echoUser();
```
jthibault03: I would suggest having more methods than just echoUser() in your Eln class, in order to organize your code a little more

src/main/java/Eln.java
jthibault03: good use of the "happy path" as outlined in the code organization guidelines, since the unique inputs (e.g. bye) are handled in the beginning. would be good to handle error cases around this section too.

src/main/java/Eln.java
```diff
@@ -0,0 +1,105 @@
+
+import java.util.Scanner;
+
+
+public class Eln {
+    private static final String LINE = "____________________________________________________________";
+    private static Task[] tasksList = new Task[100]; // max number of tasks set to 100
```
Musfirahe0556596: Plural form should be used on names representing a collection of objects! Also, a suggestion: can put the number 100 as a named constant :]

src/main/java/Eln.java
```diff
@@ -0,0 +1,105 @@
...
+
+        if (taskToAdd.startsWith("todo")) {
+            newTask = new Todo(taskToAdd);
+        } else if (taskToAdd.startsWith("deadline")) {
+            newTask = new Deadline(taskToAdd);
+        } else if (taskToAdd.startsWith("event")) {
+            newTask = new Event(taskToAdd);
+        } else {
+            newTask = new Task(taskToAdd);
+        }
```
Musfirahe0556596: might want to be careful with putting this as the default option/else loop. A task can either be a todo, deadline or event - but if user inputs some other string, a (general) Task is added which does not fall in any of the task types mentioned!

Musfirahe0556596 (commented):
Code seems oki (do implement your increments!)! Just some comments to take note of!

PR #103, opened 2024-02-05
[{Jackie Neo}] iP

JackieNeoCEG:


src/main/java/Bart.java
```diff
@@ -0,0 +1,86 @@
+import java.util.Scanner;
+
+public class Bart {
+    private static final String LINE = "____________________________________________________________";
+    private static final Task[] tasksArray = new Task[100];
+    private static int taskCount = 0;
+
+    public static void main(String[] args) {
+        greetUser();
```
CerealMiller: This is good naming sense for methods.

src/main/java/Bart.java
```diff
@@ -0,0 +1,86 @@
+import java.util.Scanner;
+
+public class Bart {
+    private static final String LINE = "____________________________________________________________";
```
CerealMiller: Very good. This is a neat way to print out lines in CLT later on.

LINE is correctly named as a constant.

src/main/java/Task.java
```diff
@@ -0,0 +1,22 @@
+// Use a Task class to represent tasks.
+public class Task {
+    protected String description;
+    protected boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
```
CerealMiller, line 47: Class is correctly named using PascalCase.

CerealMiller (commented):
Overall, the format has been good and readable.

src/main/java/Task.java
```diff
@@ -0,0 +1,22 @@
...
+
+    public String getStatusIcon() {
+        return (isDone ? "[X]" : "[ ]"); // mark done task with [X] and undone with [ ]
```
yuki-zmstr: Nice use of ternary expressions!

src/main/java/Bart.java
```diff
@@ -0,0 +1,86 @@
...
+                    tasksArray[taskCount] = new Task(command); // Create new Task object
+                    taskCount++;
+                    System.out.println(LINE);
+                    System.out.println("added: " + command);
+                    System.out.println(LINE);
+                } else {
+                    System.out.println("Sorry, task list is full.");
+                }
+            }
+        }
```
yuki-zmstr, line 63: Agreed! 

In the future, if the primary function in the if clause gets long, you can consider setting the exception case (the task list being full) to the top as a guard clause.

src/main/java/Bart.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+    public static void listTasks() {
+        System.out.println(LINE);
+        //Edge case: If list empty
```
yuki-zmstr: Nice work in setting aside the edge case.

Just a (really) minor aesthetic edit 
```suggestion
        // Edge case: If list empty
```

src/main/java/Bart.java
yuki-zmstr: Agreed. You could also consider factoring out the whole printing function as a separate function, if there are cases that you want to print multiple lines.

src/main/java/Bart.java
CerealMiller, line 63: Good format for nested while, if-else statements.

src/main/java/Bart.java
```diff
@@ -0,0 +1,86 @@
...
+            System.out.println(tasksArray[taskIndex].getStatusIcon() + " " + tasksArray[taskIndex].description);
+        } else {
+            System.out.println("Invalid task number.");
+        }
+        System.out.println(LINE);
+    }
+
+    //similar to markTask
+    public static void unmarkTask(String command) {
+        int taskIndex = Integer.parseInt(command.substring(7).trim()) - 1;
```
yuki-zmstr: .substring(7) seems to be hard-coded, and assumes the user input to follow a certain format. You could consider making the starting index dynamic, while providing some visual guide for the user, like in the Contacts example we did in class.

yuki-zmstr (commented):
Nice work on using comments and following coding standards!

Some suggestions for your persusal.

src/main/java/Bart.java
```diff
@@ -0,0 +1,131 @@
...
+            switch (command) {
+                case "list":
+                    listTasks();
+                    break;
+                case "bye":
+                    break;
+                default:
+                    handleMarkAndDelete(command);
+                    break;
+            }
```
Musfirahe0556596: the switch statement should be in the following form:

```
switch (condition) {
case ABC:
    statements;
    // Fallthrough
case DEF:
    statements;
    break;
case XYZ:
    statements;
    break;
default:
    statements;
    break;
}
```

src/main/java/Bart.java
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+
+    private static void handleMarkAndDelete(String command) {
```
Musfirahe0556596: might want to consider renaming this method name because there is a addNewTask function here too (for better readability)! :] could be smtg like performTaskOperations() as an example

src/main/java/Bart.java
```diff
@@ -0,0 +1,131 @@
...
+                    tasksList.add(new Event(command));
+                } catch (IllegalArgumentException e) {
+                    System.out.println(LINE + "\nOOPS!!! The description of a event cannot be empty.\n" + LINE);
+                    return;
+                }
+                break;
+            default:
+                System.out.println(LINE + "\nOOPS!!! I'm sorry, but I don't know what that means :-(\n" + LINE);
+                return;
+        }
```
Musfirahe0556596: the switch statement should be in the following form:

```
switch (condition) {
case ABC:
    statements;
    // Fallthrough
case DEF:
    statements;
    break;
case XYZ:
    statements;
    break;
default:
    statements;
    break;
}
```

Musfirahe0556596 (commented):
Generally oki, just some minor coding quality and standard to follow! 👍

PR #106, opened 2024-02-05
[Zheng Wentao] iP

Zhengwinter:


src/main/java/Commands.java
```diff
@@ -0,0 +1,3 @@
+public enum Commands {
+    bye,,list,mark,unmark
```
runxinghuan: I think there is a redundant comma here.

src/main/java/Winter.java
```diff
@@ -0,0 +1,161 @@
...
+                status = "[ ]";
+            }
+            System.out.print(indent);
+            System.out.println(taskList[i].getOrder() + ". " + status + " " + taskList[i].getTaskName());
+        }
+        System.out.print(line);
+    }
+    // Class for Task object
+    private static class Task {
+        boolean marked;
```
runxinghuan: This boolean should be named to sound like booleans. Consider changing it to "isMarked"?

src/main/java/Deadline.java
```diff
@@ -0,0 +1,2 @@
+package PACKAGE_NAME;public class Deadline {
```
runxinghuan: I like how you named your classes, which are all in PascalCase.

src/main/java/Winter.java
```diff
@@ -0,0 +1,161 @@
...
+            System.out.print(line);
+            System.out.print(indent);
+            System.out.println(echoLine);
+            System.out.print(line);
+            echoLine = input.nextLine();
+        }
+
+    }
+    // Method for adding tasks by handling user commands
+    private static void addTasks() {
```
runxinghuan: I think plural form is used on names representing a collection of objects only. Maybe you should use singular form here?

src/main/java/Winter.java
```diff
@@ -0,0 +1,161 @@
...
+            case "list":
+                displayList(taskList,taskIndex);
+                continue;
+            case "List":
+                displayList(taskList,taskIndex);
+                continue;
+            case "LIST":
+                displayList(taskList,taskIndex);
+                continue;
+
```
runxinghuan: I notice that you did not add a `default` case in the `switch` statement. Did you do this intentionally?

runxinghuan (commented):
Looks good to me overall, your code was easy to read. I like that you named all your classes in PascalCase and methods in camelCase. Good job!

src/main/java/Winter.java
```diff
@@ -0,0 +1,161 @@
...
+        System.out.print(line);
+    }
+    // Method for farewell message
+    private static void sayBye() {
+        String line = "-----------------------------------\n";
+        String farewell = "Farewell. Hope to see you again soon!";
+        System.out.print(line);
+        System.out.println(farewell);
+        System.out.print(line);
+    }
```
lordgareth10: I like how the naming conventions are followed well! The method names use verbs for actions like sayHi and sayBye which is compliant with the coding standards.

src/main/java/Winter.java
```diff
@@ -0,0 +1,161 @@
...
+        while (true) {
+            if(echoLine.equals("bye") || echoLine.equals("Bye") || echoLine.equals("BYE")) {
+                break;
+            }
+            System.out.print(line);
+            System.out.print(indent);
+            System.out.println(echoLine);
+            System.out.print(line);
+            echoLine = input.nextLine();
+        }
```
lordgareth10: The use of while(true) could pose some readability issues as it is not clear when the loop will terminate. You could consider refactoring the code to check for the if condition inside the conditional part of the while loop.

src/main/java/Winter.java
```diff
@@ -0,0 +1,161 @@
...
+        Task[] taskList = new Task[100];
+        int taskIndex = 0;
+        Scanner input = new Scanner(System.in);
+        while (true) {
+            String command = input.nextLine();
+            Task newTask = new Task(taskIndex+1,false,command);
+            taskList[taskIndex] = newTask;
+
+            switch(command) {
+            // Cases include farewell and list commands
```
lordgareth10: I like how there is documentation to tell the reader what the case break block is doing here!

src/main/java/Winter.java
```diff
@@ -0,0 +1,161 @@
...
+
+            }
+            switch(command.substring(0,command.indexOf(" "))) {
+            // Cases for marking tasks
+            case "mark":
+                taskList[Integer.parseInt(command.substring(5,command.length()))-1].mark();
+                continue;
+            case "unmark":
+                taskList[Integer.parseInt(command.substring(7,command.length()))-1].unmark();
+                continue;
```
lordgareth10: You could consider using an extra variable to assign the parsed input to as this will make it more obvious what exactly the method mark() is being used on. For example, 
parsedIndex = Integer.parseInt(command.substring(5,command.length()))-1;
taskList[parsedIndex].mark();

lordgareth10 (commented):
Overall, I found the code easy to read and understand due to the addition of comments in various places. Classes were named in PascalCase, and variables/methods in camelCase. However there were some readability issues which I have pointed out and gave suggestions on how to fix.

src/main/java/winter/Manager.java
```diff
@@ -0,0 +1,302 @@
...
+import static winter.Commands.*;
+
+
+public class Manager {
+    private static final String line = "-----------------------------------\n";
+    private static final String indent = "   ";
+    private static int taskIndex = 0;
+    private static final Task[] taskList = new Task[100];
+
+    public static void acceptInput () {
```
Jonoans: Try to avoid long methods like this

src/main/java/winter/Manager.java
```diff
@@ -0,0 +1,302 @@
...
+        Scanner input = new Scanner(System.in);
+        String inputString;
+        Commands action;
+        boolean flag = true;
```
Jonoans: Maybe `isFlag` would better conform to our coding standards

src/main/java/winter/Manager.java
```diff
@@ -0,0 +1,302 @@
...
+        for (int i = 0; i < taskIndex; i++) {
+            System.out.print(indent);
+            if (taskList[i].getType().equals("D")) {
+                System.out.println(taskList[i].getOrder()+1 + ". [D] " + taskList[i].doneCheckbox + " "
+                        + taskList[i].getTaskName() + " (by: " + taskList[i].getEndTime() +")");
+            }else if (taskList[i].getType().equals("E")){
+                System.out.println(taskList[i].getOrder()+1 + ". [E] " + taskList[i].doneCheckbox + " "
+                        + taskList[i].getTaskName() + " (from: " + taskList[i].getStartTime()
+                        + " to: " + taskList[i].getEndTime() + ")");
+            }else {
```
Jonoans: Missing spaces here

src/main/java/winter/task/Event.java
```diff
@@ -0,0 +1,32 @@
+package winter.task;
+
+public class Event extends Task {
+    private static final String indent = "   ";
+    protected String startTime;
+    protected String endTime;
+
+    public Event (int order,boolean isMarked, String eventName, String startTime, String endTime) {
```
Jonoans, line 12: Do pay mind to the spaces

src/main/java/winter/Manager.java
```diff
@@ -0,0 +1,302 @@
...
+                continue;
+            case BYE:
+                sayBye();
+                flag = false;
+                break;
+            case LIST:
+                displayList(taskList, taskIndex);
+                continue;
+            case MARK:
+                taskList[Integer.parseInt(inputString.substring(5)) - 1].mark();
```
Jonoans: Avoid magic numbers like this as it make not be immediately apparent what `5` means

src/main/java/winter/task/Task.java
```diff
@@ -0,0 +1,66 @@
...
+        this.doneCheckbox = "[X]";
+        System.out.println("Woohoo! I've marked this task as done:");
+        System.out.print(indent);
+        System.out.println(this.doneCheckbox + this.taskName);
+        System.out.print(line);
+    }
+    // Method for unmarking a class
+    public void unmark() {
+        String line = "-----------------------------------\n";
+        String indent = "   ";
```
Jonoans, line 43: Maybe these strings could be declared at the class-level as constants?

src/main/java/winter/task/Task.java
```diff
@@ -0,0 +1,66 @@
+package winter.task;
+public class Task {
+    boolean isMarked;
+    int order;
+    String taskName;
+
+    public String doneCheckbox ="[ ]";
```
Jonoans: This seems to be repeated in line 31 as well, maybe consider using a constant variable to store `"[ ]"` and have doneCheckbox reference that constant for its initial value?

src/main/java/winter/Manager.java
```diff
@@ -0,0 +1,302 @@
...
+    private static Commands classifyCommand(String inputString) {
+        switch (inputString) {
+        // Cases include farewell and list commands
+        case "bye":
+        case "Bye":
+        case "BYE":
+            return BYE;
+        case "list":
+        case "List":
+        case "LIST":
```
Jonoans: Understand what you are trying to do here, though maybe using `"some_string".toLowerCase()` can help you normalise the command word so you would not have to account for that many edge cases? This is outside of code quality / coding standard, just a tip that is hopefully helpful

Jonoans (commented):
Nice job with the functionality, however, do try to break long methods in your program into more digestible units. Try to avoid magic strings and magic numbers as well.

PR #144, opened 2024-02-05
[Sim Justin] iP

1simjustin:


src/main/java/Duke.java
```diff
@@ -1,10 +1,173 @@
+import java.util.Scanner;
+
 public class Duke {
```
ZMinghuiZ: it would be good if the class name can be changed to ur chatbot's name

src/main/java/List.java
```diff
@@ -0,0 +1,163 @@
...
+            this.addItemError();
+            return;
+        }
+
+        String taskName, taskStart, taskEnd;
+        boolean isValid = true;
+
+        try {
+            switch (type) {
+            case "T":
```
ZMinghuiZ: Naming of the case can be more descriptive :)

src/main/java/List.java
```diff
@@ -0,0 +1,163 @@
+import java.util.ArrayList;
+
+public class List {
```
ZMinghuiZ: Class name can be more descriptive since this class mainly is about task list

src/main/java/Deadline.java
```diff
@@ -0,0 +1,27 @@
...
+
+    /**
+     * Prints the task and its status in one line.
+     * Overwrites toString method in Java Object class.
+     * 
+     * @param None
+     * @return String representation of the task.
+     */
+    public String toString() {
+        return "[" + this.type + "]" + "[" + (this.isDone ? "X" : " ") + "] " + this.name
```
ZMinghuiZ: can use String.format to shorten the expression

ZMinghuiZ (commented):
Good job! nice documentation and exception control👍

src/main/java/List.java
```diff
@@ -91,44 +107,57 @@ public void addItem(String taskName) {
      * @return None
      */
     public void addItem(String description, String type) {
```
JeffinsonDarmawan: Really love the fact that you have not written in an arrowhead style code, but this method looks very long. Perhaps this might pose as a challenge for when you are trying to debug this method.

src/main/java/uwunzhe/commands/MarkUnmarkCommand.java
```diff
@@ -0,0 +1,40 @@
...
+    public void execute(TaskList taskList, Storage storage)
+            throws UwunzheException {
+        try {
+            int index = Integer.parseInt(taskString) - 1;
+            taskList.setItemStatus(this.commandString, index);
+
+            storage.saveData(taskList);
+
+        } catch (IndexOutOfBoundsException e) {
+            throw new UwunzheException("Huhhhhhhh? I cannot find!");
```
wwweert123: Can define error message constants at the top of class to prevent use of magic literals

src/main/java/uwunzhe/handler/Parser.java
```diff
@@ -0,0 +1,65 @@
...
+     * @return Object containing details of command to be executed.
+     * @throws UwunzheException
+     */
+    public Command parseCommand (String command, String taskString)
+            throws UwunzheException {
+        switch (command) {
+        case "list":
+            return new ListCommand(command, taskString);
+
+        case "mark":
```
wwweert123, line 44: use a "//fallthrough" comment to indicate a case without break statement

wwweert123 (commented):
Amazing code, highly OOPed, almost no coding standard violations, very upz

PR #145, opened 2024-02-05
[Rex Yong Jin Xiang] iP

rexyyong:


src/main/java/List.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public List() {
+        this.tasks = new String[100];
+    }
+
+    public void getAllTasks() {
+        int serialNumber = 1;
+        for (int i = 0; i < totalTasks; i++) {
+            System.out.print("     ");
+            System.out.println(serialNumber + ". " + tasks[i]);
```
ChongXern: Can consider adding the spaces together with the println for better readability as you have earlier

src/main/java/Omoh.java
```diff
@@ -16,10 +16,8 @@ public static void main(String[] args) {
     }
 
     //Method that prints horizontal line using "_" char
-    private static void printHorizontalLine() {
-        for (int j = 0; j < 4; j++) {
-            System.out.print(" ");
-        }
+    public static void printHorizontalLine() {
+        System.out.print("    ");
```
ChongXern: Great job in including spaces to make code more user friendly. Could avoid DRY rule in by possibly including it in your class as a part of a method

src/main/java/Task.java
```diff
@@ -0,0 +1,32 @@
+public class Task {
```
ChongXern: Good coding standards applied overall.

src/main/java/List.java
```diff
@@ -31,6 +31,24 @@ public void printAddedTask() {
...
+        totalTasks++;
+        printTaskAndDeadline(extractedInfo);
+    }
+
+    public void printTaskAndDeadline(Deadline description) {
+        Omoh.printHorizontalLine();
+        System.out.print("     ");
+        System.out.println("Got it. I've added this task:");
+        System.out.print("       ");
+        System.out.println("[D][ ]" + description.description + " (by: " + description.by + ")" );
```
ChongXern: The [D] part could be added as a boolean in the method so it could be easier to adjust in the future

src/main/java/Event.java
```diff
@@ -37,4 +37,9 @@ public static void addEvent (String input) {
         Task.totalTasks++;
         Event.addEventMessage(extractedInfo);
     }
+
+    @Override
```
ChongXern: All in all, the Java coding standards used throughout have been exemplary

ChongXern (commented):
The code here is clearly well thought out in terms of Java coding standards and highly commendable to learn from. I've gained valuable insights from this code, which I highly appreciate and will learn from.

src/main/java/Omoh.java
```diff
@@ -0,0 +1,79 @@
...
+    //If bye is not typed, we check 2 conditions
+    //if list is typed, list out all tasks
+    //if mark or unmark is typed, mark or unmark the task in tasklist
+    //if none of the above, add the typed input to task list
+    //If bye is typed, function exits
+    public static void readUserInput() {
+        String line;
+        Scanner in = new Scanner(System.in);
+        line = in.nextLine();
+        while (!line.equalsIgnoreCase("bye")) {
```
chenhowy: Consider adding the `line = in.nextLine();` after the if/else instead of writing it in every case.

src/main/java/Omoh.java
```diff
@@ -0,0 +1,79 @@
...
+        System.out.println("Bye. Hope to see you again soon!");
+        printHorizontalLine();
+    }
+
+    //Method reads in what user types
+    //If bye is not typed, we check 2 conditions
+    //if list is typed, list out all tasks
+    //if mark or unmark is typed, mark or unmark the task in tasklist
+    //if none of the above, add the typed input to task list
+    //If bye is typed, function exits
```
chenhowy: Clear description of how input is handled.

src/main/java/Task.java
```diff
@@ -0,0 +1,101 @@
+import java.lang.invoke.SerializedLambda;
+
+public class Task {
+    protected static Task[] tasks;
+    public static int totalTasks = 0;
+    protected String description;
+    protected boolean isDone;
+
+    protected String type;
```
chenhowy: What is this 'type' used for? I may be wrong, but I don't see where this variable is being used.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    public String getBy() {
+        return by;
+    }
+
+    public static void printDeadline (Deadline description) {
+        Omoh.printHorizontalLine();
+        System.out.println("Got it. I've added this task:");
+        System.out.println("[D][ ] " + description.description + " (by: " + description.by + ")" );
```
chenhowy: Perhaps the .type variable can be used here to replace "[D]", allowing for easier change of type letters in the future.

src/main/java/Task.java
rexyyong: Oh yes you're not wrong, it's not used for anything anymore. I forgot to remove it but basically previously I used type to store whether the object is Task, Deadline, or Event

src/main/java/Deadline.java
rexyyong: Oh yes thats a good idea. Thank you for the insight!

src/main/java/Omoh.java
rexyyong: thanks for the input, have implemented this!

src/main/java/omoh/Omoh.java
```diff
@@ -0,0 +1,103 @@
...
+                            "Example: if list only has 2 items, dont enter beyond 2");
+                } catch (EmptyTaskNumberException e) {
+                    System.out.println("Please enter a number after mark or unmark " +
+                            "Example: mark 1");
+                }
+            } else {
+                Task.addTask(line);
+                Task.printAddedTask();
+            }
+            line = in.nextLine();
```
nichyjt: I think this is a good attempt at making the code SLAP. I think there may be room for improvement to make this section SLAP more.

Here are some guiding questions...
1. What parts of the code may be repetitive? i.e. virtually the same for each if/else block you have
2. What part of the code does a particular function that can be isolated into a function? (like printing tasks?)

src/main/java/omoh/customexceptions/EmptyTodoException.java
```diff
@@ -0,0 +1,4 @@
+package omoh.customexceptions;
+
+public class EmptyTodoException extends Exception{
```
nichyjt: Formatting error at the curly brace!

src/main/java/omoh/customexceptions/EmptyTaskNumberException.java
```diff
@@ -0,0 +1,4 @@
+package omoh.customexceptions;
+
+public class EmptyTaskNumberException extends Exception {
```
nichyjt, line 3: I am not sure why the exception class here is completely empty? Shouldn't there be some kind of functionality?

src/main/java/omoh/tasktypes/Task.java
```diff
@@ -0,0 +1,97 @@
...
+        }
+        int taskNumber = Integer.parseInt(numberString);
+        return taskNumber;
+    }
+
+
+    //method that modifies whether task is done or not done, depending on keyword mark or unmark detected
+    public static void modifyDoneState(int taskNumber, String input) {
+        //only executes if taskNumber is valid
+        if (taskNumber != -1) {
```
nichyjt: You could possibly refactor this into a guard clause to avoid deep nesting! Think about how you can apply the idea of guard clauses thoughout your codebase to make the [happy path more prominent](https://nus-cs2113-ay2324s2.github.io/website/se-book-adapted/chapters/codeQuality.html#make-the-happy-path-prominent).

Also, there are many magic literals throughout the codebase. Do refactor them! For example, it is not entirely clear what `-1` means here.

nichyjt (commented):
Hi Rex,

Your code is generally well structured. Good job!

There are some other specific comments below. Try and fix them before the next tutorial!

For your consideration: There is a way to autoformat your code automatically every time you Ctrl-S on intellij which will save you much pain. :)

PR #150, opened 2024-02-05
[lordgareth10] iP

lordgareth10:


src/main/java/List.java
```diff
@@ -0,0 +1,48 @@
+public class List {
+    private Task[] tasks;
```
owx0130: I like that the array specifier is attached to the type `Task`, following the coding standards closely!

src/main/java/Dross.java
```diff
@@ -0,0 +1,85 @@
+import java.util.Scanner;
```
owx0130: I like that this statement only imports `Scanner` instead of doing `import java.util.*`

src/main/java/List.java
```diff
@@ -0,0 +1,48 @@
...
+        size++;
+    }
+
+    // Method to print the last added task
+    public void printLastTask() {
+        if (size > 0) {
+            System.out.println("Last added task: " + tasks[size - 1]);
+        } else {
+            System.out.println("The list is empty.");
+        }
```
owx0130: This is a good example of wrapping the conditionals according to the standards, even though the bodies are only 1 LOC

src/main/java/Task.java
```diff
@@ -0,0 +1,36 @@
+public class Task {
+    private String description;
+    private boolean isCompleted;
```
owx0130: I like that the boolean `isCompleted` was named to sound like a boolean!

src/main/java/Task.java
```diff
@@ -0,0 +1,36 @@
...
+        this.isCompleted = false;
+    }
+
+    // Get the task description
+    public String getDescription() {
+        return this.description;
+    }
+
+    // Check if the task is completed
+    public boolean isCompleted() {
```
owx0130: The method name `isCompleted` should be a verb instead. Since this is a getter, perhaps consider an alternative name such as `getIsCompleted`?

src/main/java/Task.java
```diff
@@ -0,0 +1,36 @@
...
+
+    // Mark the task as completed
+    public void completeTask() {
+        this.isCompleted = true;
+    }
+
+    // Mark the task as not completed (or "uncheck" the task)
+    public void uncheckTask() {
+        this.isCompleted = false;
+    }
```
owx0130: `completeTask` and `uncheckTask` are both setters for the boolean `isCompleted`, perhaps the method names can be changed to make them sound similar and also adhere to coding standards for boolean setters?

src/main/java/Dross.java
```diff
@@ -0,0 +1,85 @@
...
+            printLine();
+            System.out.println(line);
+            printLine();
+            line = in.nextLine();
+        }
+        goodbye();
+    }
+
+    //Method that prints bye message
+    public static void goodbye(){
```
owx0130: The method name `goodbye` should be a verb. Since this method just consists of print statements, perhaps consider an alternative name like `printGoodbyeMessage`?

src/main/java/Dross.java
```diff
@@ -0,0 +1,85 @@
...
+    //Method that prints line using "_" char
+    public static void printLine() {
+        for (int i = 0; i < 55; i++) {
+            System.out.print("_");
+        }
+        System.out.println();
+    }
+
+    //Method that echoes back whatever user typed in provide that it is not "bye"
+    public static void echo(){
```
owx0130: Although `echo` here is a verb, perhaps it could be named to be more descriptive? An alternative name could be `echoUserInput`.

owx0130 (commented):
Looks good to me, coding standards were generally well followed. Classes were named in PascalCase, and variables/methods in camelCase. There were a few minor violations to the coding standards which I have recommended solutions for in my comments.

src/main/java/Task.java
paturikarthik: Yes, I agree as well. Some level of similarity in the setters would also make it easy for code readability.

src/main/java/List.java
```diff
@@ -0,0 +1,61 @@
...
+    public void printAllTasks() {
+        if (size > 0) {
+            System.out.println("All tasks:");
+            for (int i = 0; i < size; i++) {
+                Task currentTask = tasks[i]; // Assuming tasks[i] is a Task object
+                System.out.print((i + 1) + ".");
+                System.out.println(currentTask);
+            }
+        }
+        else {
```
paturikarthik: the else statement should be in line 56, with reference to coding standards. However, this is just a one-time error, and it is not repeated elsewhere :)

src/main/java/Dross.java
```diff
@@ -0,0 +1,87 @@
...
+    //Method to list all tasks
+    public static void listAllTasks() {
+        printLine();
+        drossTaskList.printAllTasks();
+        printLine();
+    }
+
+    //Method to toggle tasks as marked or unmarked
+    public static void toggleMark(String instruction) {
+        String[] tokens = instruction.split(" ");
```
paturikarthik: Good eye for details! Plural form was used for the array names consistently throughout.

src/main/java/Dross.java
```diff
@@ -0,0 +1,87 @@
...
+            String[] parts = line.substring("event".length()).trim().split(" /from ", 2);
+            String[] timeParts = parts[1].split(" /to ", 2);
+            drossTaskList.addTask(parts[0], timeParts[0], timeParts[1]);
+        }
+        drossTaskList.printLastTask();
+        printLine();
+    }
+
+    //Method to determine what to do based on the input entered
+    public static void receiveUserInput() {
```
paturikarthik: Indentation is on point!

src/main/java/List.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+    }
+
+    //Method to toggle check task as done by index
+    public void markDoneByIndex(int index){
+        tasks[index-1].completeTask();
+    }
+
+    //Method to toggle check task as undone by index
+    public void markUndoneByIndex(int index){
```
paturikarthik: Slight nitpick from me. Although egyptian style brackets were used throughout, having a space between the function parameters and curly brace could make the code look even neater.

paturikarthik (commented):
LGTM! Overall, very easy code to read. neat!

src/main/java/Dross.java
```diff
@@ -0,0 +1,108 @@
...
+    public static String readLine() {
+        return in.nextLine().trim();
+    }
+
+    //Method to exit Dross bot
+    public static void goodbye() {
+        printLine();
+        System.out.println("Bye. Hope to see you again soon!");
+        printLine();
+    }
```
Jai2501: A better method name would be `printGoodbyeMessage`, as this improves code readability.

src/main/java/Dross.java
```diff
@@ -0,0 +1,108 @@
...
+            } catch (ArrayIndexOutOfBoundsException e) {
+                System.out.println("An event without a name, a start or end... What do you think you are, a philosopher?? event [name] /from [time1] /to [time2] is the way for goodness sakes...");
+            }
+        }
+        drossTaskList.printLastTask();
+        printLine();
+    }
+
+    //Method to determine what to do based on the input entered
+    public static void receiveUserInput() {
```
Jai2501: A better method name would be `readAndParseUserInput`, as this improves code readability.

src/main/java/Dross.java
```diff
@@ -0,0 +1,108 @@
...
+            } else if (line.startsWith("todo") || line.startsWith("deadline") || line.startsWith("event")) {
+                handleTaskCreation(line);
+            } else {
+                printLine();
+                System.out.println("Please enter a valid command");
+                printLine();
+            }
+            line = readLine();
+        }
+        goodbye();
```
Jai2501: You may also choose to split reading and parsing into different methods.

src/main/java/drosstasks/List.java
```diff
@@ -0,0 +1,68 @@
+package drosstasks;
+import drosstasks.Event;
+import myexceptions.InvalidTodoException;
+
+public class List {
+    private Task[] tasks;
+    private int size;
+
+    public List() {
+        this.tasks = new Task[100]; // initial capacity
```
Jai2501: Please avoid the use of Magic Numbers. Instead declare a constant with the value you deem is right, and use that constant.

Or better, you may choose to use `ArrayList` which dynamically resizes as per need, and you may not require such constants. However, everything comes at a tradeoff, so you may do some research into `Array` vs `ArrayList` and choose what fits the best for your needs.

src/main/java/drosstasks/List.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    //Method to toggle check task as done by index
+    public void markDoneByIndex(int index){
+        tasks[index-1].completeTask();
+    }
+
+    //Method to toggle check task as undone by index
+    public void markUndoneByIndex(int index){
+        tasks[index-1].uncheckTask();
```
Jai2501: This code is fragile and perhaps you may want to add a check for index, i.e, the value is not negative and is not greater than the size of the Array, else exceptions may be thrown.

src/main/java/myexceptions/InvalidTodoException.java
```diff
@@ -0,0 +1,4 @@
+package myexceptions;
+
+public class InvalidTodoException extends Exception{
+}
```
Jai2501: Empty Class. Please add at least a constructor with valid instantiations/assignments, so if this error is thrown, the description shown is correct.

Else please add a comment explaining when would this error be thrown and why is it empty and nothing initialised.

Jai2501 (commented):
Hi Gareth!

Thanks!

PR #160, opened 2024-02-06
[Phan Dinh Hung] iP

PDHung1104:


src/main/java/Duke.java
```diff
@@ -1,10 +1,20 @@
...
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        String name = "Stella";
+        System.out.println("Hello! I'm " + name);
+        System.out.println("What can I do for you?");
+        System.out.println("-----------------------------------");
```
ZhangWenyue3325: can make this line a void function so that the code looks less messy as this line appears many times

src/main/java/Task.java
```diff
@@ -0,0 +1,60 @@
...
+        System.out.println("-----------------------------------");
+        for (int i = 0; i < tasksCount; i++){
+            System.out.print((i + 1) + ".");
+            System.out.print("[");
+            if (tasks[i].isDone){
+                System.out.print("X");
+            } else {
+                System.out.print(" ");
+            }
+            System.out.print("] ");
```
ZhangWenyue3325: suggestions:
if (tasks[i].isDone){
    System.out.print("[X]");
} else {
    System.out.print("[ ]");
}

src/main/java/Task.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    public static void response(String command){
+        if (!command.equals("list")) {
+            String[] commandWords = command.split(" ");
+            if (commandWords[0].equals("mark")) {
+                System.out.println("-----------------------------------");
+                System.out.println("Nice! I've marked this task as done:");
+                int index = Integer.parseInt(commandWords[1]);
+                System.out.println("[X] " + tasks[index - 1].name);
```
ZhangWenyue3325: suggestions:
int index = Integer.parseInt(commandWords[1]) - 1;
then can use tasks[index] directly instead of tasks[index - 1] in the following lines

src/main/java/Duke.java
```diff
@@ -1,10 +1,20 @@
+import java.util.Scanner;
+
 public class Duke {
```
ZhangWenyue3325, line 5: maybe can change the name Duke to Stella

src/main/java/Task.java
```diff
@@ -0,0 +1,60 @@
...
+                int index = Integer.parseInt(commandWords[1]);
+                System.out.println("[ ] " + tasks[index - 1].name);
+                tasks[index - 1].isDone = false;
+            } else {
+                Task.addTask(command);
+            }
+        } else {
+            Task.listTasks();
+        }
+    }
```
ZhangWenyue3325: To many if else statements in this function, so the indentation looks messy
can consider use switch

ZhangWenyue3325 (commented):
Overall the code looks clean and neat. The commands are simple and rather short. However, the long functions such as void response are less structured. 
Good spacing!

src/main/java/Task.java
```diff
@@ -0,0 +1,60 @@
...
+            if (tasks[i].isDone){
+                System.out.print("X");
+            } else {
+                System.out.print(" ");
+            }
+            System.out.print("] ");
+            System.out.println(tasks[i].name);
+        }
+        System.out.println("-----------------------------------");
+    }
```
eliztan, line 17: You could consider adding a empty line in between methods

src/main/java/Task.java
```diff
@@ -0,0 +1,60 @@
...
+
+    private static int tasksCount = 0;
+    private static Task[] tasks = new Task[100];
+
+    Task(String name){
+        this.name = name;
+        isDone = false;
+    }
+
+    public static void listTasks(){
```
eliztan: Good job following camel case for method names

src/main/java/Task.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+    public static void addTask(String taskName){
+        System.out.println("-----------------------------------");
+        System.out.println("added: " + taskName);
+        tasks[tasksCount] = new Task(taskName);
+        tasksCount++;
+        System.out.println("-----------------------------------");
+    }
+
+    public static void response(String command){
```
eliztan: "command" is a slightly ambiguous name especially as it is a string of multiple words, maybe consider changing it to a clearer name

src/main/java/Task.java
```diff
@@ -0,0 +1,60 @@
...
+        if (!command.equals("list")) {
+            String[] commandWords = command.split(" ");
+            if (commandWords[0].equals("mark")) {
+                System.out.println("-----------------------------------");
+                System.out.println("Nice! I've marked this task as done:");
+                int index = Integer.parseInt(commandWords[1]);
+                System.out.println("[X] " + tasks[index - 1].name);
+                System.out.println("-----------------------------------");
+                tasks[index - 1].isDone = true;
+            } else if (commandWords[0].equals("unmark")){
```
eliztan: You could consider adding a space in between the ) and { at the end

src/main/java/Parser.java
```diff
@@ -0,0 +1,156 @@
+public class Parser {
+    private static void checkValidDeadline (String command) throws WrongInputFormat, MissingEntries, MissingTaskName {
+        String[] commandWords = command.split("/");
+
+        if (commandWords.length != 2) {
+            throw new WrongInputFormat();
+        } else {
+            if (commandWords[0].length() < 10) {
```
wenxin-c, line 14: Try not to use magic literals e.g. magic numbers and magic string, consider creating a constant and use the constant in your code instead.

src/main/java/Parser.java
```diff
@@ -0,0 +1,156 @@
...
+            return false;
+        } catch (MissingEntries me) {
+            System.out.println(UI.LINE_SEPARATOR);
+            System.out.println("Missing deadline");
+            System.out.println(UI.LINE_SEPARATOR);
+            return false;
+        } catch (MissingTaskName mtn) {
+            System.out.println(UI.LINE_SEPARATOR);
+            System.out.println("Missing task name");
+            System.out.println(UI.LINE_SEPARATOR);
```
wenxin-c: Can consider abstracting this further to a printing method that takes in a string argument to avoid repeating yourself.

src/main/java/Task.java
```diff
@@ -0,0 +1,127 @@
...
+        System.out.println(UI.LINE_SEPARATOR);
+
+        switch (task.getTaskType()) {
+            case DEADLINE:
+            case EVENT:
+            case TODO:
+                System.out.println("Got it. I've added this task:");
+                tasks.add(task);
+                break;
+        }
```
wenxin-c: Do follow the coding standard for switch block on our module website, i.e. ```switch``` and ```case``` should be aligned, there's no extra indentation in front of ```case```.

src/main/java/Task.java
```diff
@@ -0,0 +1,127 @@
...
+        System.out.println(UI.LINE_SEPARATOR);
+    }
+
+    private static boolean isValidTask (String command) {
+        return command.equals("todo") || command.equals("deadline") || command.equals("event");
+    }
+
+    private static boolean isValidCommand (String command) {
+        return command.equals("list") || command.equals("mark") || command.equals("unmark") || command.equals("delete")
+                || isValidTask(command);
```
wenxin-c: This is expression is very long and complex, try to simplify by breaking it down further.

wenxin-c (commented):
Well done in your ip so far. Try to be more careful of minor details and I would recommend to abstract your code further, especially the printing part to avoid repeating yourself.

PR #177, opened 2024-02-06
[Ranj Nicoli M. Negapatan] iP

nicknamenic:


src/main/java/Nocturne.java
```diff
@@ -0,0 +1,94 @@
+import java.util.Scanner;
+
+public class Nocturne {
+    public static void main(String[] args) throws NocturneException {
+        Task[] tasks = new Task[100];
```
Musfirahe0556596: just a recommendation: perhaps you might want to put the number 100 as a named constant to avoid magic numbers! it will also be easier for you to find & change this number in the future if everything is put on top :]

src/main/java/Nocturne.java
```diff
@@ -0,0 +1,94 @@
...
+        int taskCount = 0;
+
+        String input;
+        Scanner in = new Scanner(System.in);
+        System.out.println("Good evening. I'm Nocturne.");
+        System.out.println("What ails you on this fine day?");
+
+        input = in.nextLine();
+        while (!input.equals("bye")) {
+            String[] commandCheck = input.split(" ");
```
Musfirahe0556596: note that plural form should be used on names representing a collection of objects. might want to change this to a different name instead to for better readability - perhaps something like commandParts/commandTypes or other appropriate names!

src/main/java/Nocturne.java
```diff
@@ -0,0 +1,94 @@
...
+                    trueEvent = new Event(eventName, from, to);
+                    tasks[taskCount] = trueEvent;
+                    taskCount++;
+                    System.out.println("An event I see. I have added it:");
+                    System.out.println("  " + trueEvent);
+                    break;
+
+                default:
+                    throw new NocturneException("Your command is invalid. Try again.");
+            }
```
Musfirahe0556596: switch statement should be in the following form:

```
switch (condition) {
case ABC:
    statements;
    // Fallthrough
case DEF:
    statements;
    break;
case XYZ:
    statements;
    break;
default:
    statements;
    break;
}
```

Musfirahe0556596: additionally, might want to consider moving & refactoring the different operations to different methods (e.g. listTasks(),  markTask()) so that everything is not under the main method! try to avoid long methods :]

src/main/java/Nocturne.java
```diff
@@ -0,0 +1,94 @@
...
+                        throw new NocturneException("Your list is either empty or your brain is.");
+                    }
+                    int listIndex = Integer.parseInt(commandCheck[1]);
+                    System.out.println("Do not neglect your duties. I have marked this task as unfinished:");
+                    tasks[listIndex - 1].isDone = false;
+                    System.out.println("  " + tasks[listIndex - 1]);
+                    break;
+                }
+                case "deadline":
+                    String[] deadlineSeparated = input.split("/");
```
Musfirahe0556596: Plural form should be used on names representing a collection of objects

src/main/java/Nocturne.java
```diff
@@ -0,0 +1,94 @@
...
+                    String todoName = input.substring(5);
+                    Todo trueTodo = new Todo(todoName);
+                    tasks[taskCount] = trueTodo;
+                    taskCount++;
+                    System.out.println("A Todo task I see. I have added it:");
+                    System.out.println("  " + trueTodo);
+                    break;
+
+                case "event":
+                    String[] eventSeparated = input.split("/");
```
Musfirahe0556596: Plural form should be used on names representing a collection of objects.

Musfirahe0556596 (commented):
Generally oki job but some coding standard and quality to follow! do avoid long methods 👍

PR #182, opened 2024-02-07
[Yao Chenguang] iP

classskipper351:


src/main/java/Qchat.java
```diff
@@ -0,0 +1,62 @@
...
+       return command;
+
+
+
+   }
+
+    private static void echo(String command) {
+        System.out.print("---------------------------------------------\n");
+        System.out.print(command +"\n");
+        System.out.print("---------------------------------------------\n");
```
e0958902, line 36: I would suggest using System.lineSeparator() instead of "\n" for code compatibility with other systems.

src/main/java/Todolist.java
```diff
@@ -0,0 +1,131 @@
+import java.util.Scanner;
+
+public class Todolist {
+    private static final int MAX_LIST_LENGTH = 100;
+    static String[] todolist = new String[MAX_LIST_LENGTH];
+    static boolean[] isDoneList = new boolean[MAX_LIST_LENGTH];
```
e0958902: Good use of a static final int, instead of specifying "100" directly in the array as you have avoided using magic numbers.

src/main/java/Qchat.java
```diff
@@ -0,0 +1,62 @@
+import javax.swing.*;
```
e0958902, line 1: Would it be better to not import all packages although according to google it is lightweight.

src/main/java/Qchat.java
```diff
@@ -0,0 +1,62 @@
...
+                   +"knows better about life and better about you\n"
+                   +"What can I do for you?\n"
+                   +"----------------------------------------------------------------\n") ;
+   static final String GOODBYE_GREETING = new String(
+           "--------------------------------------------------------\n" +
+            "goodbye\n"+ "Qchat, your life-long trusted companion\n");
+
+
+
+
```
e0958902: You may want to reduce the number of line breaks.

src/main/java/Todolist.java
```diff
@@ -0,0 +1,131 @@
...
+
+    public static void PrintList(){
+        int i=1;
+        System.out.print("---------------------------------------------\n");
+        System.out.println("Here are your current tasks in your list:");
+        for (String s :todolist ){
+            if(s == null){
+                continue;
+            }
+            System.out.printf("%d."+GetIcon(i)+s+"\n",i);
```
e0958902: You may want to add spaced before and after the "+", this is part of coding standards.

src/main/java/Listmanager.java
```diff
@@ -0,0 +1,133 @@
+import java.util.Scanner;
+
+public class Todolist {
+
+    String Icon = "T";
+    private static final int MAX_LIST_LENGTH = 100;
```
ChoonSiang: Good use of constant for the max list length

src/main/java/Listmanager.java
```diff
@@ -0,0 +1,133 @@
...
+    static String[] todolist = new String[MAX_LIST_LENGTH];
+    static boolean[] isDoneList = new boolean[MAX_LIST_LENGTH];
+
+    static int ListCount= 0;
+    static Scanner in = new Scanner(System.in);
+
+    public static void AddToList(){
```
ChoonSiang:
```suggestion
    public static void AddToList() {
```

Take note that for function declaration, there should be a space before {

src/main/java/Listmanager.java
```diff
@@ -0,0 +1,133 @@
...
+
+
+    static void HandleList(){
+        PrintList();
+        String command ="";
+
+        while(!command.equals("quit")){
+            command = in.nextLine();
+            switch (command){
+            case"add":
```
ChoonSiang: Inconsistency in whether there is a space after case.
```suggestion
            case "add":
```

src/main/java/Listmanager.java
```diff
@@ -0,0 +1,133 @@
...
+                break;
+            }
+
+        }
+
+    }
+
+    public static void PrintList(){
+        int i=1;
+        System.out.print("---------------------------------------------\n");
```
ChoonSiang: Consider making this a constant to prevent repetition of the same code and will be more consistent when modification is made

src/main/java/Qchat.java
```diff
@@ -0,0 +1,62 @@
...
+       case "list":
+           Todolist.HandleList();
+           break;
+
+
+       default:
+           echo(command);
+           break;
+
+       }
```
ChoonSiang: Why is the other commands such as quit, help, mark, unmark etc. only accessible after the user input list?

src/main/java/Listmanager.java
```diff
@@ -0,0 +1,133 @@
...
+                PrintList();
+                break;
+            case"quit":
+                System.out.println("Task change complete");
+                break;
+            case"help":
+                System.out.print("supported commands: add , quit , help , mark,unmark ,clear, list\n");
+                break;
+            case "clear":
+                ClearList();
```
ChoonSiang: Good attempt at abstracting the functionality of each command

src/main/java/Listmanager.java
```diff
@@ -0,0 +1,133 @@
...
+        System.out.println("please type the task number you want to set as done \n");
+        PrintList();
+        int number= in.nextInt();
+
+        while(number>ListCount){
+            System.out.println("task do not exist, pleas enter a valid task number ");
+            number= in.nextInt();
+        }
+
+        isDoneList[number-1]=true;
```
ChoonSiang:
```suggestion
        isDoneList[number-1] = true;
```

src/main/java/Listmanager.java
```diff
@@ -0,0 +1,133 @@
...
+                continue;
+            }
+            System.out.printf("%d."+ GetIcon(i)+s+"\n",i);
+            i++;
+        }
+        System.out.print("---------------------------------------------\n");
+    }
+    static void ClearList(){
+        int i;
+        for (i=0;i<MAX_LIST_LENGTH;i++){
```
ChoonSiang:
```suggestion
        for (i = 0; i < MAX_LIST_LENGTH; i++) {
```

ChoonSiang (commented):
Good attempt at abstraction, but need to work on formatting, especially on the usage of space to follow the coding standard and allow for easier readability. Also need to consider the logic of your program whereby the rest of the commands is only accessible after the user input list.

PR #190, opened 2024-02-08
[Huang Maodian] iP

Geinzit:


src/main/java/huan/main/Huan.java
```diff
@@ -0,0 +1,142 @@
+package huan.main;
+
+import huan.task.*;
```
ChoonSiang: Avoid importing *, try to explicitly list out the import classes

src/main/java/huan/main/Huan.java
```diff
@@ -0,0 +1,142 @@
...
+                DeadlineTask deadlineTask = (DeadlineTask)task;
+                deadlineTask.printTask();
+                break;
+            default:
+                task.printTask();
+                break;
+            }
+        }
+    }
+    public static void readProcessCommand() {
```
ChoonSiang: Very long method, maybe try to abstract each command?

src/main/java/huan/main/Huan.java
```diff
@@ -0,0 +1,142 @@
...
+
+        Scanner scanner = new Scanner(System.in);
+        while(true) {
+            System.out.println("-------------------------");
+            String inputCommand = scanner.nextLine();
+            String[] words = inputCommand.split(" ");
+            String firstWord = words[0];
+            String suffixWord;
+            if(words.length > 1) {
+                suffixWord = inputCommand.substring(words[0].length() + 1);
```
ChoonSiang: Multiple usage of Magic numbers, try to assign the number with a meaningful variable name for easy understand of what the number is for.

src/main/java/huan/task/DeadlineTask.java
```diff
@@ -0,0 +1,53 @@
...
+
+public class DeadlineTask extends Task{
+    private String ddlTime;
+    public DeadlineTask(String nameWithDate, Boolean isDone) throws Exception{
+        StringBuilder ddlTime = new StringBuilder();
+        StringBuilder name = new StringBuilder();
+        String[] words = nameWithDate.split(" ");
+        /*
+          state:
+          0 means currently concatenating name
```
ChoonSiang: Can consider using ENUM instead of 0,1 for the state

src/main/java/huan/task/DeadlineTask.java
```diff
@@ -0,0 +1,53 @@
+package huan.task;
+
+import java.util.Objects;
+
+public class DeadlineTask extends Task{
+    private String ddlTime;
```
ChoonSiang, line 12: maybe can consider calling it deadlineTime, as ddlTime can be hard to understand what it means

src/main/java/huan/task/EventTask.java
```diff
@@ -0,0 +1,68 @@
...
+                case (1):
+                    startTime.append((startTime.length() == 0 ? "" : " ")).append(word);
+                    break;
+                case (2):
+                    endTime.append((endTime.length() == 0 ? "" : " ")).append(word);
+                    break;
+                }
+            }
+        }
+        if(state != 2 || name.toString().isEmpty() || startTime.toString().isEmpty() || endTime.toString().isEmpty()) {
```
ChoonSiang: very long and complicated statement, can consider splitting it up

ChoonSiang (commented):
Generally ok, but try to be more careful with magic number and string. Avoid long and complicated function/statement.

PR #192, opened 2024-02-08
[Liu Ze Hui] iP

liuzehui03:
Added Level-0, Level-1, Level-2, Level-3


src/main/java/Task.java
```diff
@@ -0,0 +1,32 @@
+public class Task {
+    protected static String[] tasks = new String[100];
+    protected static boolean[] tasksCompleted = new boolean[100];
```
yeozhishen: whichTasksCompleted sounds like a better variable name as the variable name tasksCompleted sounds like an integer

yeozhishen (commented):
LGTM! Your code mostly seems to follow the coding standards, all but one line appears to need changing!

src/main/java/Duke.java
```diff
@@ -1,10 +1,58 @@
...
+//so scanner class can be used
+
 public class Duke {
-    public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
+    public static void printLine() {
```
yeozhishen: Should try to rename it to something that indicates it is also printing a newline ie)printlnLine

src/main/java/Duke.java
```diff
@@ -1,10 +1,58 @@
...
+        System.out.println("________________________________________");
+    }
+    public static void printShortLine() {
```
yeozhishen: Should try to rename it to something that indicates it is also printing a newline ie)printlnShortLine

src/main/java/Duke.java
```diff
@@ -1,10 +1,58 @@
...
+        String userInput;
+        userInput = in.nextLine();
+
+        while(!userInput.equals("bye")) {
+            printShortLine();
+            System.out.println("added: " + userInput);
+            printLine();
+            if (userInput.equals("list")) {
+                Task.printList();
+            } else if(userInput.matches("mark \\d+")) {
```
yeozhishen: may want to set constants to identify the regex expression, as regex may not be intuitively readable

src/main/java/Duke.java
```diff
@@ -1,10 +1,58 @@
...
+                String[] taskIndex = userInput.split(" ");
+                int index = Integer.parseInt(taskIndex[1]);
+                Task.markDone(index-1);
+                System.out.println("Nice! I've marked this task as done:");
+                System.out.println("  [X] " + Task.tasks[index-1]);
+            } else if(userInput.matches("unmark \\d+")) {
```
yeozhishen: may want to set constants to identify the regex expression, as regex may not be intuitively readable

yeozhishen (commented):
suggested extra comments for cleaner code

src/main/java/Duke.java
```diff
@@ -1,10 +1,58 @@
+import java.util.Scanner;
+//so scanner class can be used
+
 public class Duke {
```
runxinghuan: Consider changing the class name to `Yoj`? Since you named your bot to be this.

src/main/java/Duke.java
```diff
@@ -1,10 +1,58 @@
...
+
+        while(!userInput.equals("bye")) {
+            printShortLine();
+            System.out.println("added: " + userInput);
+            printLine();
+            if (userInput.equals("list")) {
+                Task.printList();
+            } else if(userInput.matches("mark \\d+")) {
+                String[] taskIndex = userInput.split(" ");
+                int index = Integer.parseInt(taskIndex[1]);
```
runxinghuan: Consider changing these two lines to a method? So that you can use again in "unmark" condition as well.

src/main/java/Duke.java
```diff
@@ -1,10 +1,58 @@
...
+                Task.markDone(index-1);
+                System.out.println("Nice! I've marked this task as done:");
+                System.out.println("  [X] " + Task.tasks[index-1]);
```
runxinghuan: Since you already have the `getStatusIcon` method, why not use it here?

src/main/java/Duke.java
```diff
@@ -1,10 +1,58 @@
...
+            } else if(userInput.matches("unmark \\d+")) {
+                String[] taskIndex = userInput.split(" ");
+                int index = Integer.parseInt(taskIndex[1]);
+                Task.markUndone(index-1);
+                System.out.println("OK, I've marked this task as not done yet:");
+                System.out.println("  [ ] " + Task.tasks[index-1]);
```
runxinghuan: Consider extracting these lines to a method called `markTask`? So that you can have the same level of abstraction within a code fragment.

src/main/java/Task.java
```diff
@@ -0,0 +1,32 @@
...
+    protected static int taskNumber;
+
+    public Task() {
+        taskNumber = 0;
+    }
+
+    public static void addTask(String userInput) {
```
runxinghuan: I like the way you named your methods, which are all in camalCase. Good job!

runxinghuan (commented):
Generally quite good. However, I noted a few minor coding quality violations which I have recommended solutions for in the comments.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,12 @@
+public class Deadline extends Task {
+    protected String by;
+
+    public Deadline (String description, String by){
+        super(description);
+        this.by = by;
+    }
```
Jai2501: Advisable to have a uniform spacing strategy, to enhance code readability.

```suggestion
    public Deadline (String description, String by) {
        super(description);
        this.by = by;
    }
```

src/main/java/Event.java
```diff
@@ -0,0 +1,18 @@
...
+    }
+
+    public String getAt(){
+        return at;
+    }
+
+    public String toString() {
+        return "[E] " + super.toString() + description + " (at: " + at + ")";
+    }
+}
```
Jai2501: It is usually a practise to use `this.method()` or `this.variable` when that `method` or `variable` belongs to the Class.

This change can be made at other places as well :)

```suggestion
    public String getAt(){
        return this.at;
    }

    public String toString() {
        return "[E] " + super.toString() + this.description + " (at: " + this.at + ")";
    }
}
```

src/main/java/Task.java
Jai2501: Please avoid the use of Magic Numbers. Instead declare a constant with the value you deem is right, and use that constant.

Or better, you may choose to use `ArrayList` which dynamically resizes as per need, and you may not require such constants. However, everything comes at a tradeoff, so you may do some research into `Array` vs `ArrayList` and choose what fits the best for your needs.

Jai2501: For this line, I may suggest you to refactor `Task` Class to have a boolean variable `isDone`. As a new `Task` is created, that `isDone` variable is unique to that task and may be updated as required. This is also efficient as you don't have to maintain an extra `Array`. 

Sometimes it may happen that, you delete a `Task` and you may forget to update it in this other `Array`, making your code fragile and more bug prone.

```Java
    protected boolean isTaskDone;
```

src/main/java/ToDo.java
```diff
@@ -0,0 +1,12 @@
...
+        super(description);
+    }
+
+    // Prints task
+    public void printTask(){
+        System.out.print("[T][");
+
+        System.out.println("] " + description);
+    }
+}
```
Jai2501: Missing EOF. Add an extra empty line at the end of file (EOF), as this indicates the end of the file.

You may configure IntelliJ to add this automatically for you when hit `CTRL/CMD + S`.

src/main/java/Yoj.java
```diff
@@ -0,0 +1,58 @@
...
+        String userInput;
+        userInput = in.nextLine();
+
+        while(!userInput.equals("bye")) {
+            printShortLine();
+            System.out.println("added: " + userInput);
+            printLine();
+            if (userInput.equals("list")) {
+                Task.printList();
+            } else if(userInput.matches("mark \\d+")) {
```
Jai2501: You may want to declare all the regex expressions being used as constants. This also allows you to just update the constant and reuse that at multiple places, without needing to update it at every place you declared it initially.

src/main/java/Yoj.java
```diff
@@ -0,0 +1,58 @@
...
+                String[] taskIndex = userInput.split(" ");
+                int index = Integer.parseInt(taskIndex[1]);
+                Task.markUndone(index-1);
+                System.out.println("OK, I've marked this task as not done yet:");
+                System.out.println("  [ ] " + Task.tasks[index-1]);
+            } else {
+                Task.addTask(userInput);
+            }
+            userInput = in.nextLine();
+        }
```
Jai2501: This code block can be abstracted to improve the code readability. Usually, methods should not be longer than 30 Lines of Code.

Reference: https://nus-cs2113-ay2324s2.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-long-methods

Jai2501 (commented):
Hi Ze Hui!

Thanks!

PR #194, opened 2024-02-09
Sparsh/IncyBot iP

IncyBot:


src/main/java/Incy.java
```diff
@@ -0,0 +1,77 @@
...
+public class Incy {
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+        String input;
+        Task[] tasks = new Task[100];
+        int taskCount = 0;
+        String logo = " ____  _  _  ___  _  _ \n"
+                    + "(_  _)( \\( )/ __)( \\/ )\n"
+                    + " _)(_  )  (( (__  \\  / \n"
+                    + "(____)(_)\\_)\\___) (__) \n";
```
Jai2501: It is good to specify access modifiers for the variables being declared, like `private`, `public` or `protected`

src/main/java/Incy.java
```diff
@@ -0,0 +1,77 @@
+import java.util.Scanner;
+
+public class Incy {
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+        String input;
+        Task[] tasks = new Task[100];
```
Jai2501: Please avoid the use of Magic Numbers. Instead declare a constant with the value you deem is right, and use that constant.

src/main/java/Incy.java
```diff
@@ -0,0 +1,77 @@
...
+                }
+            }
+        }
+
+        System.out.println("____________________________________________________________\n"
+                            + "Cya later mate!"
+                            + "\n____________________________________________________________"
+        );
+        scanner.close();
+    }
```
Jai2501: This code block is deeply nested, and needs to be worked on. 

In general, this method is too long, and it is preferred for methods to be shorter than 30 lines of code. 

Please work on abstracting this chunk of code.

References: 
1. https://nus-cs2113-ay2324s2.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-deep-nesting
2. https://nus-cs2113-ay2324s2.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-long-methods

Jai2501 (commented):
Hi Sparsh!

Thanks!

PR #195, opened 2024-02-10
[Yung Ka Shing] iP

Andy123qq4:


src/main/java/Duke.java
```diff
@@ -1,10 +1,249 @@
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Scanner;
+
 public class Duke {
```
ChoonSiang: Try to split the classes into individual class file

src/main/java/Duke.java
```diff
@@ -1,10 +1,249 @@
...
+        this.tasks = new ArrayList<>();
+    }
+
+    public void greet() {
+        System.out.println("=========================");
+        System.out.println("Hello! I'm " + name);
+        System.out.println("What can I do for you?");
+    }
+
+//    public void addTask(String description) {
```
ChoonSiang: Avoid redundant commented codes, especially when there is an existing function with the same name

src/main/java/Duke.java
```diff
@@ -1,10 +1,249 @@
...
+//        Task task = new Task(description);
+//        tasks.add(task);
+//        System.out.println("added: " + description);
+//    }
+
+    public void addTask(String description) {
+        String[] parts = description.split(" ", 2);
```
ChoonSiang: Try to avoid magic number, try to give a meaningful variable name to the number chosen for easier understanding of why the number is chosen

src/main/java/Duke.java
```diff
@@ -1,10 +1,249 @@
...
+        String taskType = parts[0].toLowerCase();
+        String taskDescription = parts.length > 1 ? parts[1] : "";
+
+        Task task;
+        switch (taskType) {
+            case "todo":
+                task = new Todo(taskDescription);
+                break;
+            case "deadline":
```
ChoonSiang: Consider abstracting out the functionality of each task type

src/main/java/Duke.java
```diff
@@ -1,10 +1,249 @@
...
+        } else {
+            System.out.println("Here are the tasks in your list:");
+            for (int i = 0; i < tasks.size(); i++) {
+//                System.out.println((i + 1) + ".[" + (tasks.get(i).isDone ? "X] " : " ] ") + tasks.get(i));
+                System.out.println((i + 1) + "." + tasks.get(i));
+            }
+        }
+    }
+
+    public void markTaskAsDone(String description) {
```
ChoonSiang: Try to avoid arrowhead looking function

src/main/java/Duke.java
```diff
@@ -1,10 +1,249 @@
...
+                break;
+            }
+        }
+
+        if (!found) {
+            System.out.println("Task not found.");
+        }
+    }
+
+    public void echoCommands() {
```
ChoonSiang: Function name does not really match what it does

src/main/java/Duke.java
```diff
@@ -1,10 +1,249 @@
...
+
+        @Override
+        public String getType() {
+            return "T";
+        }
+    }
+
+    public class Deadline extends Task {
+
+        private String ddl;
```
ChoonSiang: Avoid using variable name that can be confusing. Can just set it to deadline instead.

ChoonSiang (commented):
Generally ok with the coding standard and quality, but slightly behind on progress. Need to take note to practice more abstraction and split the classes into different files.

PR #207, opened 2024-02-16
[Hyojin Jung] ip

hailey-jung:


src/main/java/Hailey.java
```diff
@@ -0,0 +1,73 @@
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Scanner;
+
+public class Hailey {
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+        List<Task> tasks = new ArrayList<>();
```
Jai2501, line 8: It is good to specify access modifiers for the variables being declared, like `private`,`public` or `protected`.

src/main/java/Hailey.java
```diff
@@ -0,0 +1,73 @@
...
+                    System.out.println("Invalid command format.");
+                }
+            } else {
+                tasks.add(new Task(input));
+                System.out.println("added: " + input);
+            }
+        }
+
+        scanner.close();
+    }
```
Jai2501: This code block is deeply nested, and needs to be worked on.

In general, this method is too long, and it is preferred for methods to be shorter than 30 lines of code.

Please work on abstracting this chunk of code.

References:
1. https://nus-cs2113-ay2324s2.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-deep-nesting
2. https://nus-cs2113-ay2324s2.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-long-methods

Jai2501 (commented):
Hi Hailey!

Thanks!

